Issue #42, opened 2015-01-17
adding mzidentml data to MSexp object fails


adder:
Hi

I try to add a .mzid file to my MSexp object but when I look at fData I only see NA's. There are for sure identification in the file (it's 50 mb and when I open it in an editor I see lots of entries). The .mzid is generated by msgf+ on a search perfomed on a MGF file. The MSexp object is generated with the same MGF file. I also loaded the results in Peptideshaker and tried to create an .mzid file with the peptideshaker export. Same result.
When I try adding identification files in MSnbase with the files provided with the MSnbase package, it works.

Is it something to do with the MGF file?

Greetz

sgibb:
Which version of MSnbase do you use? What is the output of 'sessionInfo()'? Could you share the corresponding files? 

Am 17. Januar 2015 13:06:26 MEZ, schrieb adder <notifications@github.com>:
>Hi
>
>I try to add a .mzid file to my MSexp object but when I look at fData I
>only see NA's. There are for sure identification in the file (it's 50
>mb and when I open it in an editor I see lots of entries). The .mzid is
>generated by msgf+ on a search perfomed on a MGF file. The MSexp object
>is generated with the same MGF file. I also loaded the results in
>Peptideshaker and tried to create an .mzid file with the peptideshaker
>export. Same result.
>When I try adding identification files in MSnbase with the files
>provided with the MSnbase package, it works.
>
>Is it something to do with the MGF file?
>
>Greetz
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/lgatto/MSnbase/issues/42


Sent from my mobile device. Please excuse my brevity.

lgatto:
> Hi
>
> I try to add a .mzid file to my MSexp object but when I look at fData I only see NA's. There are for sure identification in the file (it's 50 mb and when I open it in an editor I see lots of entries). The .mzid is generated by msgf+ on a search perfomed on a MGF file. The MSexp object is generated with the same MGF file. I also loaded the results in Peptideshaker and tried to create an .mzid file with the peptideshaker export. Same result.
> When I try adding identification files in MSnbase with the files provided with the MSnbase package, it works.
>
> Is it something to do with the MGF file?

Yes, indeed. The MGF file is very metadata poor, and the matching
between the spectra in the `MSnExp` object and the identification
results can not be guaranteed. It would probably be possible to get some
code to work for specific pipelines (say Mascot), but without much
generality.

However, using files that store metadata in a well defined standard will
[...]

Issue #43, opened 2015-01-19
xtandem results in MSexp?


adder:
Hi, 
You are probably aware that xtandem does not use the mzid output format for its result. The xtandem XML can thus not be loaded directly into MSexp. It is possible to read the results with rtandem package. Is there some way to get your xtandem results  or the rtandem result class into MSexp without using external tools like http://www.psidev.info/mzidentml to convert the xml to mzid (which didn't work btw when I tried is, some nulpointerexecption)

greetz

lgatto:
> You are probably aware that xtandem does not use the mzid output format for its result. 
> The xtandem XML can thus not be loaded directly into MSexp. 

Yes, I know that. We prefer to focus our effort on community-wide standards although, to be fair, xTANDEM's format has been out there for longer.

> It is possible to read the results  with rtandem package. Is there some way to get your 
> xtandem results or the rtandem result class into MSexp without using external tools like 
> http://www.psidev.info/mzidentml to convert the xml to mzid (which didn't work btw when I 
> tried is, some nulpointerexecption)

Once you have the xTANDEM results in R, as a `data.frame` for example, and you can match the identification results and the spectra in the `MSnExp`, you will have to re-order the identification `data.frame` and then do the following to add these data to the existing feature/spectra annotation:

```r
fd <- fData(msexp) 
## merge or cbind fd and your id data, keeping
## the original row order; for example using dplyr::left_join
## fd <- left_join(fd, id)
fData(msexp) <- fd
``` 

(You might want to send a bug report re the conversion error.)

sgibb:
I am not familiar with the `rTANDEM` package. Do we get a `data.frame` with all necessary information? If this is the case we could think about a function similar to our current `addIdentificationData` method that accept a `data.frame` as input. That should be not too much work because we currently convert mzIdentML into a `data.frame` and do the merging via `utils.mergeSpectraAndIdentificationData` (hidden by `NAMESPACE`). `utils.mergeSpectraAndIdentificationData` needs both the feature and identification data as `data.frame` and returns a merged feature `data.frame`.

lgatto:
`rTANDEM`'s output is a bit convoluted, IMHO. It would be possible to get a peptide `data.frame` by first extracting all proteins, then all peptides. But I haven't used it enough to make sure this would work for sure and that the matching with the original data, whether `mgf` or `mzML` would be straightforward. 

[...]

PR #45, opened 2015-01-19
fix mzIdentML import for MGF based identification files; see issue #42

sgibb:
This PR is WIP. @lgatto's suggestion https://github.com/lgatto/MSnbase/issues/42#issuecomment-70466709 and the documentation are still missing. I will leave a comment if this PR is finished.


sgibb:
This PR may breaks the API because the generic for `addIdentificationData` was changed:
```r
# before PR
setGeneric("addIdentificationData", function(object, ...) standardGeneric("addIdentificationData"))
# PR
setGeneric("addIdentificationData", function(object, id, ...) standardGeneric("addIdentificationData"))
```
Additionally the name of the second argument was `filenames` before (is `id` now).

`idSummary` is currently not working (because `mzIDCollection` contains no hint about the corresponding id files; see https://github.com/thomasp85/mzID/pull/27), some tests fail (all regarding the identification file problem) and the documentation is still missing.

sgibb:
@lgatto a few questions:

1. Do we really want to add the filenames of the identification files to `MSnExp@MSnProcess@files` slot (via `fileNames(msexp) <- basename(mzID::files(id)$raw$location)`)? This would not work for the general `data.frame` if it contains no "idFile" column (that's why I removed it from the current code in this PR, but it could be added easily).
2. By adding the new overloaded methods `addIdentificationData` we break the backward compatibility. How we want to deal with this fact (the second argument is now `id` instead of `filenames`)?

sgibb:
A 3. questions same problem as 2.: backward compatibility:
The merge columns are now `fcol = c("spectrum.file", "acquisition.number"), icol = c("spectrumFile", "acquisitionnum")` (was `fcol = c("file", "acquisition.number"), icol = c("spectrumFile", "acquisitionnum")` before but it wasn't visible to the user). 

Both `fcol` columns are generated from the `MSnExp` as follows:
```r
fd$spectrum.file <- basename(fileNames(object)[fromFile(object)])
fd$acquisition.number <- acquisitionNum(object)
```
The `icol` columns are present in the `mzID` output. IMHO it is easier for the user to create his own `spectrumFile` and `acquisitionNum` columns (if he want to create his own `data.frame` and don't use `mzID`) than creating a `file` column (file id; position of the file in the `MSnExp@MSnProcess@files` slot). The latter could be done automatically for `mzID` but the user has to do it himself for his own `data.frame`s.

This change breaks the backward compatibility. Should not be a real problem because nobody could add his own `data.frame` yet. (But if somebody works with the `fData(msexp)$identFile` column he will have trouble.)

The output of `idSummary` would change as well. Instead of file id numbers there would be filenames.

What do you think?

lgatto:
> Do we really want to add the filenames of the identification files to MSnExp@MSnProcess@files slot (via fileNames(msexp) <- basename(mzID::files(id)$raw$location))? This would not work for the general data.frame if it contains no "idFile" column (that's why I removed it from the current code in this PR, but it could be added easily).

Would it be possible to just ignore the identification filename if there is no `idFile` column? We don't want this to become overly complicated, so I would also consider dropping it altogether.

lgatto:
> By adding the new overloaded methods `addIdentificationData` we break the backward compatibility. How we want to deal with this fact (the second argument is now `id` instead of `filenames`)?

Yes, but I think we gain much from this. There have been 2 or 3 issues about adding id data, so some flexibility is welcome. In the future, I would also consider add a method for identification from `mzR`.

lgatto:
> A 3. questions same problem as 2.: backward compatibility:

Same as for 2 - go ahead, break it ;-)

sgibb:
IMHO this PR is ready to merge if `mzID` 1.5.2 is on bioc. Therefore @thomasp85 has to export his `mzID` and `mzIDCollection` classes or accept this PR: https://github.com/thomasp85/mzID/pull/28

The identification filenames are ignored now (was issue 1.). The backward compatibility is broken (were 2. and 3.).

#42 is solved with this PR by the following clumsy call:
```r
library("MSnbase")

msexp <- readMgfData("3_2.mgf")
msexp <- addIdentificationData(msexp, "3_2.msgf.mzid",
                               fcol=c("spectrum.file", "TITLE"),
                               icol=c("spectrumFile", "spectrum title"))
```

Theoretical #43 should be solved as well but I don't know the correct columns.

thomasp85:
Will do tomorrow - lot on my plate atm.

lgatto:
Sorry, dump question here
```r
msexp <- addIdentificationData(msexp, "3_2.msgf.mzid",
                               fcol=c("spectrum.file", "TITLE"),
                               icol=c("spectrumFile", "spectrum title"))
```
but why do you have `spectrum.file/spectrumFile` and `TITLE/spectrum title`? Is the matching not done on one only?

sgibb:
In this particular case `spectrum.file/spectrumFile` is indeed superfluous. The following would be enough:
```r
msexp <- addIdentificationData(msexp, "3_2.msgf.mzid",
                               fcol=c("TITLE"),
                               icol=c("spectrum title"))
```
I had multiple file support in mind. But I have seen that `readMgfData` doesn't support multiple files. That's why it should be safe to use just `TITLE/spectrum title`.

`mzID` 1.5.2 is on bioc; If you agree this PR could be merged.

lgatto:
> If you agree this PR could be merged.

Sure, go ahead!

lgatto:
FYI: I have now pushed the latest version to the Bioc svn server.

PR #48, opened 2015-02-01
mzR as backend for addIdentificationData

sgibb:
Dear Laurent,

this PR introduces `mzR` as backend for `addIdentificationData`.

Sadly it is far from perfect. While I could circumvent most problems by renaming the columns of the `mzR::psms` output similar to the output of `flatten(mzID(...))` the label-free quantification fails because `mzR::psms` has no `length` column. Is there a way to calculate this column? Otherwise we need to open an issue against `mzR`.

```r
quantFile <- dir(system.file(package = "MSnbase", dir = "extdata"),
                 full.name = TRUE, pattern = "mzXML$")
identFile <- dir(system.file(package = "MSnbase", dir = "extdata"),
                 full.name = TRUE, pattern = "dummyiTRAQ.mzid")

msexp <- readMSData(quantFile)

msexpMzID <- addIdentificationData(msexp, identFile, backend="mzID")
msexpMzR <- addIdentificationData(msexp, identFile, backend="mzR")

quantify(msexpMzID, "SIn")
quantify(msexpMzR, "SIn")
# Error in SI(object, method, ...) :
#  lengthnot found in fvarLabel(.). 'plength' must a feature variable
```

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -29,7 +29,7 @@ Author: Laurent Gatto <lg390@cam.ac.uk> with contributions from
 	Vladislav Petyuk, Thomas Naake and Sebastian Gibb.
 Maintainer: Laurent Gatto <lg390@cam.ac.uk>
 Depends: R (>= 3.1), methods, BiocGenerics (>= 0.7.1),
-	 Biobase (>= 2.15.2), mzR, BiocParallel
+	 Biobase (>= 2.15.2), mzR (>= 2.1.6), BiocParallel
 Imports: plyr, IRanges, preprocessCore, vsn, grid,
 	 reshape2, stats4, affy, impute, pcaMethods,
 	 MALDIquant (>= 1.9.8), digest, lattice, ggplot2
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -32,10 +32,9 @@ Depends: R (>= 3.1), methods, BiocGenerics (>= 0.7.1),
 	 Biobase (>= 2.15.2), mzR, BiocParallel
 Imports: plyr, IRanges, preprocessCore, vsn, grid,
 	 reshape2, stats4, affy, impute, pcaMethods,
-	 mzID (>= 1.5.2), MALDIquant (>= 1.9.8), digest,
-	 lattice, ggplot2
+	 MALDIquant (>= 1.9.8), digest, lattice, ggplot2
 Suggests: testthat, zoo, knitr (>= 1.1.0), rols, Rdisop, pRoloc,
-	  pRolocdata (>= 1.0.7), msdata, roxygen2, rgl, BiocStyle
+	  pRolocdata (>= 1.0.7), msdata, roxygen2, rgl, BiocStyle, mzID (>= 1.5.2)
 License: Artistic-2.0
 LazyData: yes
 VignetteBuilder: knitr
--- a/NAMESPACE
+++ b/NAMESPACE
@@ -22,8 +22,6 @@ importFrom(vsn, vsn2, meanSdPlot)
 importFrom(IRanges, IRanges)
 importClassesFrom(IRanges, IRanges)
 importFrom(affy, MAplot, ma.plot, mva.pairs)
-importFrom(mzID, mzID, flatten)
-importClassesFrom(mzID, mzID)
 importFrom(digest, digest)
 
 export(MSnSet,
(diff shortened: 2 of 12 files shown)


lgatto:
I would suggest to open an `mzR` issue to add a protein length column.

sgibb:
Thanks to your `mzR` fixes this code works just fine now. Currently I replace some column names of the `mzR::psms` output with the corresponding `mzID::flatten` names. If `mzR` will be our new default backend it would more reasonable to do it the other way around (replace the `mzID` column names with their `mzR` counterparts). That would be important because some functions, e.g. `quantify`, need some specific column names, e.g. `length`(mzID)/`DBseqLength`(mzR).

https://github.com/lgatto/MSnbase/pull/48/files#diff-cbb7235afee10c3646161358a7655b0aR725

lgatto:
Thanks! Let's wait a bit before switching completely to `mzR`, as there still seems to be issues with `score` for example.

sgibb:
Because of the definition of `addIdentificationData(..., backend = c("mzR", "mzID"), ...)` `mzR` is already the default backend. Should we swap both backends?

lgatto:
I think depending on the backends, the feature variables are named differently, which made the `removeNoId` example to fail. I have changed it now, so let's it like that, or do you think it is better to swap?

sgibb:
If `mzR` is our preferred backend we should translate all `mzID::flatten` column names into `mzR` compatible ones. Also we need to change all methods that depend on column names (e.g. `summaryId`, `removeNoId`, ...).

In general we have to find a way to handle this. IMHO the following solutions are possible:

1. We translate `mzID` column names into `mzR` column names.
2. We encourage the authors of `mzID` and `mzR` to find a consens and use the same column names for the `mzID::flatten` and `mzR::psms/mzR::score` output.
3. We drop the `mzID` support and switch to `mzR` completely.

I would welcome solution 2. but to be honest I don't believe it is possible. The easiest and most consistent way for most users (and us developers) would be 3.

lgatto:
> 1. We translate `mzID` column names into `mzR` column names.
> 2. We encourage the authors of `mzID` and `mzR` to find a consens and use the same column names for the `mzID::flatten` and `mzR::psms`/`mzR::score` output. 

The column names are generated by different software (pure `R` or through `Rcpp`). One would need to understand the different conventions to, then, translate one into the other.

>3. We drop the `mzID` support and switch to `mzR` completely.

I have modified the `removeNoId` example recently, as it failed due to different names, as you described. Wouldn't it be less work to keep `mzID` as the default back end for now? I don't think `MSnbase` currently requires the faster `mzR` access.  Making sure that things work with either `mzR` or `mzID` and keep the other back end as backup is fine by me.

sgibb:
Ok, so I will switch back to `mzID`. I am going to translate most (if not all) `mzR` columns to `mzID::flatten` names to ensure that all our functions work.

lgatto:
I did not mean to translate all col names to `mzID` names and yet use `mzR` (I meant just keep using the former). I think we should use the names used in the backend that is parsing the id data (too confusing otherwise and more difficult to maintain). Let's go for `mzR`, then. Unless it's a simple revert to a commit, point me to the code chunk and I will change to col names translation. Sorry for the confusion.

sgibb:
Is there any serious problem with `mzR`? Or any other reason why we should not switch to `mzR` completely. I can switch back to `mzID` but except `addIdentificationData` no other function will work with the `mzR` output out of the box (the user has to find the column name himself; functions with hard-coded columns like `idSummary` won't work for `mzR` without translating the column names). IMHO it would be easier to support just one mzIdentML-package.
Sorry that I don't think about that before asking for merging this PR.

lgatto:
> IMHO it would be easier to support just one mzIdentML-package.

Yes, I agree, and it should be whatever is the easiest. 
(`MSnbase` will never have to open hundreds of `mzid` files, so `mzID`, even if slower, should be fine - as opposed to`Pbase`).

sgibb:
Ok, I completely revert this PR from master. I hope everything went well and it won't have any sideeffects. So there should not be any `mzR` mzIdentML support anymore. `mzID` is now our only mzIdentML backend (as it was before).

I am very sorry that I caused so much confusion with this PR.

lgatto:
No worries. Support for `mzR` will be added during the next major re-factoring of the package.

PR #50, opened 2015-04-14
fix of combine features bug

vladpetyuk:
fix of bug during combine features for one sample with "multiple" redundancy handling


lgatto:
Thanks!

PR #51, opened 2015-05-10
MALDIquant 1.12

sgibb:
Dear Laurent,

the upcoming MALDIquant release will have some changes to internal functions. Currently the only one that affects MSnbase is `.estimateNoise` that returns a simple numeric vector instead of a matrix.

This PR is a reminder. I will merge it myself when MALDIquant 1.12 is released on CRAN (I will need some more days to test etc.). But it would be necessary to cherry-pick this PR into the stable/release branch on bioc as well.

Best wishes,

Sebastian


lgatto:
Thanks for the note.

PR #58, opened 2015-06-25
Added description in iPQF header

martifis:
I edited the description in the iPQF.R from the iPQF branch. I just realised that this iPQF.R version I forked from your repository was missing 'the examples and reference' on the iPQF branch, these are only found in the iPQF.R on the master. This confused me a bit and I stayed with the branch.

So far, I only edited the description, but maybe adding some more information about the features used in the details section might be helpful after their naming is solved? Also I could add sth why a combination with MedianPolish might be useful.

### DIFF
--- a/R/iPQF.R
+++ b/R/iPQF.R
@@ -188,10 +188,13 @@ uni.measured.dist <- function(pos, uniques.all, mat) {
 
 
 ##' The iPQF spectra-to-protein summarisation method integrates
-##' peptide characteristics and quantitative values for protein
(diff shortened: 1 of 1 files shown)


lgatto:
Thanks. I will delete the `iPQF` branch now to avoid any confusoin. `iPQF` was a so called feature branch, were I worked on a specific new feature and, once working, got merged into `master`, which is the main branch, and has now been further improved.
I have copied you improved description and added your paper as *in prep*.

PR #64, opened 2015-09-24
  use split in subsetBy

sgibb:
Dear Laurent,

this PR was described in #63 and replaces the use of `lapply` by `split`. Unfortunately it changes the order of the output (but as you mentioned in #63 the order should not be important; BTW you could specify the order of levels in a `factor` by setting the levels manually: `factor(c("A", "B", "C"), levels=c("B", "A", "C"))`).

Additionally I changed the order in the unit test (row `1:3` are now `c(2, 3, 1)`).

Best wishes,

Sebastian

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -322,27 +322,20 @@ getTopIdx <- function(X, n, fun, ...) {
 }
 
 subsetBy <- function(X, groups, byIdx) {
-  ans <- c()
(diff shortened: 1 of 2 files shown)


lgatto:
Let's wait after next release to merge this.

lgatto:
@sgibb could you confirm that this PR is still relevant and valid - if so, I'll merge.

sgibb:
It is still relevant and valid.

PR #72, opened 2015-12-14
apply functions columnwise by group

sgibb:
This PR introduces `utils.applyColumnwiseByGroup` a function similar to `tapply` but works on matrices and applies functions like `colSums` groupwise.

While looking into `nQuants` I recognized that it does exactly the same as `featureCV` (except one line => the workhorse function) but with different and complicated code. I removed the duplicated code and create a new function `utils.applyColumnwiseByGroup` instead.

Changes:
- `featureCV` adapted to `utils.applyColumnwiseByGroup`
- `nQuants` adapted to `utils.applyColumnwiseByGroup`
- unit test for `utils.applyColumnwiseByGroup`

I also recognized that `nQuants` arguments are completely different from `topN`, `combineFeatures` etc. That's why I replaced the `fcol` argument by the `groupBy` argument that is present in all other `MSnSet` functions. This **breaks the API**! Should I revert it?

IHMO a lot of more functions could use the `utils.applyColumnwiseByGroup` function. E.g. the `combineMatrixFeatures` could use it instead of its `by(...); do.call(rbind, ...)` approach.
Should I try to convert them as well?

BTW: I don't like the name `utils.applyColumnwiseByGroup`.

### DIFF
--- a/R/functions-MSnSet.R
+++ b/R/functions-MSnSet.R
@@ -118,53 +118,41 @@ updateFeatureNames <- function(object, label, sep = ".") {
 ##' This function is typically used after \code{\link{topN}} and before
 ##' \code{\link{combineFeatures}}, when the summerising function is
 ##' \code{sum}, or any function that does not normalise to the number of
-##' features aggregated. In the former case, sums of feautres might
+##' features aggregated. In the former case, sums of features might
 ##' be the result of 0 (if no feature was quantified) to \code{n}
 ##' (if all \code{topN}'s \code{n} features were quantified) features,
 ##' and one might want to rescale the sums based on the number of
 ##' non-NA features effectively summed.
 ##'
 ##' @title Count the number of quantitfied features.
-##' @param object An instance of class \code{"\linkS4class{MSnSet}"}.
-##' @param fcol The feature variable to consider when counting the
-##' number of quantified featues.
+##' @param x An instance of class \code{"\linkS4class{MSnSet}"}.
+##' @param groupBy An object of class \code{factor} defining how to summerise the features.
+##' @return A \code{matrix} of dimensions \code{length(levels(groupBy))} by \code{ncol(x)}
 ##' @return A \code{matrix} of dimensions
 ##' \code{length(levels(factor(fData(object)[, fcol])))} by \code{ncol(object)}
 ##' of integers.
-##' @author Laurent Gatto
+##' @author Laurent Gatto <lg390@@cam.ac.uk>,
+##' Sebastian Gibb <mail@@sebastiangibb.de>
 ##' @examples
 ##' data(msnset)
 ##' n <- 2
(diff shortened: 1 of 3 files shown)


lgatto:
Are `groupBy` and `fcol` really the same? It looks like the former is a factor that is passed explicitly by the user 
```r
featureCV
function (x, groupBy, na.rm = TRUE, norm = c("sum", "max", "none", 
    "center.mean", "center.median", "quantiles", "quantiles.robust")) 
{
    groupBy <- as.factor(groupBy)
    norm <- match.arg(norm)
    if (norm != "none") 
        x <- normalise(x, method = norm)
    j <- split(1L:nrow(x), groupBy)
    ans <- matrix(NA_real_, nrow = nlevels(groupBy), ncol = ncol(x), 
        dimnames = list(levels(groupBy), colnames(x)))
    for (i in seq(along = j)) {
        subexprs <- exprs(x)[j[[i]], , drop = FALSE]
        ans[i, ] <- utils.colSd(subexprs, na.rm = na.rm)/colMeans(subexprs, 
            na.rm = na.rm)
    }
    colnames(ans) <- paste("CV", colnames(ans), sep = ".")
    ans
}
```
while the latter is a character that refers to a feature variable (checked with `fcol %in% fvarLables(object)` below)
```r
> nQuants
function (object, fcol) 
{
    .count <- function(x) {
        m <- rep(nrow(x), ncol(x))
        nna <- apply(x, 2, function(.x) sum(is.na(.x)))
        m - nna
    }
    if (class(object) != "MSnSet") 
        stop("'object' must be of class 'MSnSet'.")
    if (missing(fcol)) 
        stop("'fcol' is required.")
    if (!fcol %in% fvarLabels(object)) 
        stop("'fcol' not found in fvarLabels(object).")
    res <- by(exprs(object), factor(fData(object)[, fcol]), .count)
    if (ncol(object) == 1) {
        ans <- as.matrix(res)
    }
    else {
        ans <- do.call(rbind, res)
    }
    colnames(ans) <- sampleNames(object)
    return(ans)
}
```

(I agree that this is not consistent, and should probably be addressed).

sgibb:
> (I agree that this is not consistent, and should probably be addressed).
That's why I changed it. IMHO there is no reason why `nQuants` should be different from `topN`.

lgatto:
@sgibb this slip under the radar. Feel free to merge when you have time.

lgatto:
@sgibb I would suggest that next time, we add/modify a function parameter, we keep the old one for backwards compatibility - something like

```r
nQuants <- function (x, groupBy, fcol = groupBy) {
  if (!missing(fcol)) 
    ## a message and proper handling
   ...
}
```

sgibb:
You are absolutely right! Sorry.
Should we reintroduce fcol for the next bioc release?

lgatto:
No, let's leave it like it is now - I have documented the change in the man page (see MSnbase commit 16f3bb69ed1f27452f4d1cf32bae40b8cab530fa).

PR #84, opened 2016-04-24
replace fixed width by IRanges::findOverlaps in removePrecMz

sgibb:
I was wondering if we could remove the fixed width for a precursor by an approach that removes the complete peak regardless of its width.

Instead of `which(object$mz > precMz[1] & object$mz < precMz[2])` I used `MSnbase:::relaxedMatch` but `MSnbase:::matchPeaks(..., method="highest")` would be even better but requires converting of the list objects in `removePrecMz_list` into `Spectrum` objects (or a rewrite of `matchPeaks`).

``` r
library("git2r")
library("devtools")

repo <- repository()

checkout(repo, "master")
load_all()
data(itraqdata)
sp <- itraqdata[[1]]
ms <- utils.removePrecMz(sp)

checkout(repo, "removePrecMz")
load_all()
pr <- utils.removePrecMz_Spectrum(sp)

png("removePrecMz.png")
par(mfcol=c(1, 2))
plot(mz(sp), intensity(sp), type="b",
     xlim=c(519, 523), ylim=c(0, 8e5), main="master")
lines(mz(ms), intensity(ms), type="b", col=2, pch=3)
plot(mz(sp), intensity(sp), type="b",
     xlim=c(519, 523), ylim=c(0, 8e5), main="PR")
lines(mz(pr), intensity(pr), type="b", col=3, pch=4)
dev.off()
```

removeprecmz
(black: original spectrum; red/green: after removal of the precursor)

If we want to remove the whole region around the precursor this won't work.

What do you think? 
@lgatto maybe you could ask the author Guangchuang Yu (I am not sure whether he has an github account).


lgatto:
This is an excellent suggestion - yesterday I thought about promoting `removePrezMz` to a generic and have methods for `MSnExp` and `Spectrum`. Actually it might be useful to also have a `removePeak` method, to have a general interface, but with the same implementation.

Note that there's already a `removePeaks` (that remove peaks lower than a threshold) and `removeReporters` (that removes reporter ions). The latter could use your implementation. 

Regarding the old behaviour, we could either have an argument, or, better, a new method `removeMzRegion`. 

What do you think?

lgatto:
Summary of the above:
- `removePeaks(object, t)`: sets peaks < `t` from individual spectra or whole experiments to 0; keep as is. 
- `removePeak(object, mz)`: using new implementation, removes a specific peak.
- `removePrecMz(object)`: as above, but uses `precursorMz` to get `mz`.
- `removeReporters(object)`: as above, but uses a `ReporterIons` to get the `mz` values.
- `removeMzRegion(object, mz)`: sets all values between `mz[1]` and `mz[2]` to 0.

lgatto:
@sgibb - FYI, I plan to look into this in more details and implement my suggested updates after release.

lgatto:
I forgot about this, but would like to include it. @sgibb - anything else before I address conflicts to merge?

sgibb:
I think it is save to merge this. JFTR: as mentioned above I use `relaxedMatch` here which always looks for the closest match. In general the `precMz` should be closest one to the highest intensity in the range of the precursor peak but if not this method could fail. It would be better to use `matchPeaks(..., method="highest")`. `matchPeaks` needs a `Spectrum` as input. So we have to rewrite `matchPeaks` or `removePrecMz` (or add some additional functions) to handle these edge cases.

Are you still want to implement `remove{Peak[s],PrecMz,Reporters,MzRegion}`?

lgatto:
@sgibb - yes, I would still like to generalise this functionality to `remove{Peak[s],PrecMz,Reporters,MzRegion}`. I have opened #252.

Could you have a look at the conflict in `utils.R`. I have the feeling that we want to keep the [current `utils.clean`](https://github.com/lgatto/MSnbase/blob/master/R/utils.R#L130-L157) rather than the [version in the `removePrecMz` branch](https://github.com/lgatto/MSnbase/blob/885941470874052b9f6e6399b99e9f8caef952fa/R/utils.R#L111-L136) to be merged.

sgibb:
Yes, you are absolutely right. I merged master into the `removePrecMz` branch. Unfortunately (but expected) the unit tests for `removePrecMz` fail (missing `precMz` and `width` argument).

To be honest I never completely tested this function. It was meant as suggestion and minimal working example. So there would be definitively some more work to do before we could merge it into master.

lgatto:
Ok, thanks, I will try to make some progress on this today.

lgatto:
It looks like things can be fixed by using `MSnbase:::utils.removePrecMz_Spectrum` rather than `MSnbase:::utils.removePrecMz`, which is the function that we should use anyway. Does this make sense to you?

sgibb:
> Does this make sense to you?
Yes, `MSnbase:::utils.removePrecMz_Spectrum` should be used.

PR #85, opened 2016-05-04
Fix wrong indexing in readMSdata, msLevel==1

jotsetung:
Dear Laurent,
I suspect there is a wrong indexing in line 180 in readMSData.R (`fullhdorder[fullhdorder] <- .fname` should be `fullhdorder[fullhdordercounter] <- .fname`). Sorry for all the other differences which seem to be introduced by my emacs (deleting all trailing whitespaces).

cheers, jo


lgatto:
Thanks!

PR #91, opened 2016-06-07
Fix bug in normalize-MSnExp

jotsetung:
o Fix a bug in normalize method for MSnExp objects: assigning normalized
  spectra directly to assayData is not possible, as the environment is
  locked.
o readMSData: if no phenodata is provided it creates an empty one with
  rownames corresponding to the file names.

### DIFF
--- a/R/functions-MSnExp.R
+++ b/R/functions-MSnExp.R
@@ -148,19 +148,22 @@ clean_MSnExp <- function(object, all, verbose = TRUE) {
 
 
 normalise_MSnExp <- function(object,method) {
-  sapply(featureNames(object),
-         function(x) {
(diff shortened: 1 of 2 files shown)


lgatto:
> Fix a bug in normalize method for MSnExp objects: assigning normalized spectra directly to assayData is not possible, as the environment is locked.

Good catch. This is actually a sneaky one. The tests did not catch this because, the data that is used doesn't lock the bindings!
```r
> data(itraqdata)
> range(intensity(itraqdata)[[1]])
[1]       0 1076030
> x <- normalise(itraqdata, method = "max")
> range(intensity(itraqdata)[[1]])
[1] 0 1
```

PR #203, opened 2017-04-07
Add $,pSet method to easily access columns in the phenoData

jotsetung:
- Add $, $<- and pData<- methods to easily get and replace/add columns to the
  phenodata.
- Add related unit tests and add documentation.

Basically, It facilitates accessing pheno data columns by just calling `pset$pheno_col`. I find that pretty convenient e.g. in `ExpressionSet` objects and alike, so I thought it might be nice to have it in `MSnbase` as well.
Also I added replace methods for the `pData`.

### DIFF
--- a/NAMESPACE
+++ b/NAMESPACE
@@ -267,7 +267,10 @@ exportMethods(updateObject,
               ## MzTab
               fileName,
               metadata,
-              psms)
+              psms,
+              pData, "pData<-",
+              "$",
+              "$<-")
 
 ## methods NOT exported
 ## curveStats
--- a/R/methods-pSet.R
+++ b/R/methods-pSet.R
@@ -536,6 +536,19 @@ setMethod("spectrapply", "pSet", function(object, FUN = NULL,
     return(vals)
 })
 
+setMethod("$", "pSet", function(x, name) {
+    eval(substitute(pData(x)$NAME_ARG, list(NAME_ARG = name)))
+})
+setReplaceMethod("$", "pSet", function(x, name, value) {
+    pData(x)[[name]] <- value
+    x
+})
+
+setReplaceMethod("pData", "pSet", function(object, value) {
+    pData(object@phenoData) <- value
+    object
+})
+
 ################################
 ## TODO: setReplaceMethods for
 ##  phenoData
--- a/man/pSet-class.Rd
+++ b/man/pSet-class.Rd
@@ -10,6 +10,8 @@
 \alias{[,pSet,ANY,ANY,ANY-method}
 \alias{[[,pSet-method}
 \alias{[[,pSet,ANY,ANY-method}
+\alias{$,pSet-method}
+\alias{$<-,pSet-method}
 \alias{abstract,pSet-method}
 \alias{acquisitionNum,pSet-method}
 \alias{scanIndex,pSet-method}
@@ -44,6 +46,7 @@
 \alias{mz,pSet-method}
 \alias{notes,pSet-method}
 \alias{pData,pSet-method}
+\alias{pData<-,pSet,ANY-method}
 \alias{peaksCount,pSet,missing-method}
 \alias{peaksCount,pSet,numeric-method}
 \alias{phenoData,pSet-method}
@@ -151,6 +154,10 @@
       return object of same class. }
     \item{[[}{\code{signature(x = "pSet")}: Direct access to individual
       spectra. }
+    \item{$}{\code{signature(x = "pSet")}: directly access a specific
+    sample annotation column from the \code{pData}.}
+    \item{$<-}{\code{signature(x = "pSet")}: replace or add a
+    sample annotation column in the \code{pData}.}
     \item{abstract}{Access abstract in \code{experimentData}. }
     \item{assayData}{\code{signature(object = "pSet")}: Access the
       \code{assayData} slot. Returns an \code{environment}. }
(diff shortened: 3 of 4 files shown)


lgatto:
Thanks @jotsetung, I think this will prove very useful. I'll wait for the new release to merge. 

On a similar note, I also thought it would be useful to be able to change the behaviour of `$` through a `MSnbase` option, where it could either link to `pData` (default) or `fData`. The latter is much more useful in our spatial proteomics work in `pRoloc`, where we so often need to access feature variables in our `MSnSet`s.

jotsetung:
Don't you think that might become a little confusing if `$` is used to access both rows or columns? What about `$$` to access `fData` and `$` for `pData`?

lgatto:
Hmm, `$$`, that is an interesting suggestion. Or what about `£` (no, that's a joke) :-). Ok, let's go for `$$`, as that will even bypass the need to setting an options. Whoever has time for this is welcome to have a go, but won't be merged before the new release.

jotsetung:
I checked the `$$` option but that does not seem to be that straight forward. Didn't find where `$` is defined to check how and if a function named `$$` could be defined to have the same behavior (i.e. that it is possible to call it like `x$$y` instead of `$$(x, y)`.
@lgatto do you know by chance how and where `$` is defined?

sgibb:
I am not sure that it is possible to use `$$` as function. We can define other operators:
```r
library("MSnbase")
data(msnset)

setGeneric("%fd%", function(object, name)standardGeneric("%fd%"))
setGeneric("%pd%", function(object, name)standardGeneric("%pd%"))

setMethod("%fd%", "MSnSet", function(object, name)fData(object)[, as.character(substitute(name))])
setMethod("%pd%", "MSnSet", function(object, name)pData(object)[, as.character(substitute(name))])

msnset %fd% ProteinAccession
#  [1] BSA     ECA1422 ECA4030 ECA3882 ECA1364 ECA0871
#  [7] ECA4512 ECA4513 ECA3969 ECA3082 ECA1032 ECA1294
# [13] ECA4514 ECA1104 ECA3356 ECA4037 ECA0621 ECA1093
# [19] ECA0452 ENO     ECA2391 ECA0435 ECA1362 ECA1363
# [25] ECA3566 ECA0435 ECA0691 ECA3566 ECA3377 ECA0978
# [31] ECA2831 ECA4514 ECA0469 ECA4514 ECA4514 ECA0172
# [37] ECA0631 ECA3349 ECA0469 ECA4514 ENO     ENO
# [43] ECA4026 ECA3929 ECA4514 ECA4026 ECA4013 BSA
# [49] BSA     ECA2186 ENO     ECA3349 ECA2421 ECA1443
# [55] ECA3175
# 40 Levels: BSA ECA0172 ECA0435 ECA0452 ... ENO

msnset %pd% mz
# [1] 114.1 115.1 116.1 117.1
```
(`%fd<-%`, `%pd<-%` would be possible, too)
Of course *tab completion* is not working. Maybe we can hack something via `?rcompgen`. But I am not sure that is worth the effort: `msnset %fd% ProteinAccession` is IMHO no real improvement over `fData(msnset)$ProteinAccession`.

sgibb:
```r
`$$` <- function(x, y) { x[[y]] }
l <- list(a=1, b=2)
l$$b
# Error: unexpected '$' in "l$$"
$$(l, "a")
# Error: unexpected '$' in "$"
```

jotsetung:
Yes, that was what I realized too, defining `$$` is tricky.

Regarding the `%fd%` etc operators, it would be possible, but might not be intuitive. What I mean is that many users that use e.g. `ExpressionSet`, `SummarizedExperiment` etc are familiar with the use of `$` to access phenodata columns. That was also the first thing I missed in the `pSet`. I would not implement a completely new shortcut to do the same (i.e. the `%pd%`).

Eventually we just add the `$` and use the old fashioned `fData(msd)` to access feature annotations?

lgatto:
Have a look in [`Biobase`](https://github.com/Bioconductor-mirror/Biobase/blob/master/R/methods-eSet.R#L404-L416) to see how things have been implemented for `$`.

jotsetung:
That's exactly our implementation for the `$` in `MSnbase`. The problem is that `$$` does not work in a similar fashion (as @sgibb pointed out).

lgatto:
Following up from @sgibb's example:
```r
[1] 1
> `$$` <- function(x, y) { x[[y]] }
> l <- list(a=1, b=2)
> `$`(l, "a")
[1] 1
> `$$`(l, "a")
[1] 1
```
I think the problem we have is that R can't parse `$$` in `l$$a`, while it knows how to handle `$`.

lgatto:
An alternative would be to use `@`, but I always stumble on the wrong signature
```r
> .Foo <- setClass("Foo", slots = c(a = "numeric"))
> foo <- .Foo(a = 1:10)
> names(foo@a) <- letters[1:10]
> `$`
.Primitive("$")
> getGeneric("$")
standardGeneric for "$" defined from package "base"
function (x, name) 
standardGeneric("$", .Primitive("$"))
<bytecode: 0x1355f40>
<environment: 0x1221b00>
Methods may be defined for arguments: x
Use  showMethods("$")  for currently available ones.
> setMethod("$", "Foo", function(x, name) x@a[name])
[1] "$"
> foo$a
a 
1 
```
but
```r
> getGeneric("@")
NULL
> `@`
.Primitive("@")
> setMethod("@", "Foo", function(x, name) x@a[name])
Error in setGeneric(f, where = where) : 
  ‘@’ dispatches internally;  methods can be defined, but the generic function is implicit, and cannot be changed.
> setMethod("@", "Foo", function(object, name) object@a[name])
Error in setGeneric(f, where = where) : 
  ‘@’ dispatches internally;  methods can be defined, but the generic function is implicit, and cannot be changed.
```

jotsetung:
I think `@` would also confuse people since it's the accessor for the slots of the class.

lgatto:
Yes, but sometime the perfect is the enemy of good. I am unsure that `$$` would ever work, hence may be better `@` than nothing, ... My idea or using `£` might become the least bad solution in the end!

lgatto:
I am going to merge this. Issue #209 follows up the discussion to directly access `fData` columns.

PR #204, opened 2017-04-21
Add "normalise to reference run (per protein)" function.

sgibb:
This PR should introduce another normalisation/label free quantitation method.

It is described in: https://doi.org/10.1104/pp.114.245589

> The peptide data were converted to protein intensities as follows. For each protein, a fraction with the highest number of peptides quantified was nominated as a reference fraction. The protein abundance in that reference fraction was taken as one. Then, other fractions were quantified against the reference fraction by finding the ratio of the combined intensity for peptides shared between the interrogated and reference fractions. When the quantities in all fractions for a given protein were computed, the values were renormalized to give a sum = 1 across all 10 fractions.

It was already implemented by @lgatto in https://github.com/lgatto/pRoloc/blob/master/R/lopims.R#L80-L127. And we need it in our current *synapter*/LOPIMS analysis.

I like to rewrite it (add some unit tests etc.). IMHO it is not only useful for organelle proteomics or *pRoloc* users, so I think *MSnbase* would be the right place for this method.

While having a prototype of the workhorse functions I was wondering what would be the right top-level API for such a function. It is similar to the TOP3 approach which is run in the following way:
```r
## extract top 3 peptides
msnset <- topN(msnset, groupBy=proteinAccession, n=3)

## calculate the number of peptides that are available
nPeps <- nQuants(msnset, groupBy=proteinAccession)

## sum top3 peptides into protein quantitation
msnset <- combineFeatures(msnset, groupBy=proteinAccession, fun="sum", na.rm=TRUE)

## adjust protein intensity based on actual
## number of top peptides
exprs(msnset) <- exprs(msnset) * (3/nPeps)
```
A `normToReference` function would internally convert a `MSnSet` of peptides into an `MSnSet` of proteins with its own functions. There is no simple `combineFeatures` method aka `sum` because the reference run is different for each protein and the number of summed peptides could be different for each run within each protein (so even `nQuants` would need a different implementation). So it would be more like:
```r
msnset <- normToReference(msnset, groupBy=proteinAccession)
```

So before working on a `normToReference` function I would like to ask the following questions:

1. Which kind of high-level (user-visible) API I should use/provide?
2. @lgatto Do you want to have this function in *MSnbase* (or keep it in *pRoloc*)?
3. Currently everything is in `utils.R`. Is there a better file for it?
4. Any ideas for a better name?

### DIFF
--- a/R/functions-MSnSet.R
+++ b/R/functions-MSnSet.R
@@ -33,7 +33,7 @@ normalise_MSnSet <- function(object, method, ...) {
       e <- sweep(e, 2L, cmeds - med)
   } else {
     switch(method,
-           max = div <- apply(exprs(object), 1L, max, na.rm = TRUE),
+           max = div <- .rowMaxs(exprs(object), na.rm = TRUE),
            sum = div <- rowSums(exprs(object), na.rm = TRUE))
     e <- exprs(object)/div
   }
--- a/R/utils.R
+++ b/R/utils.R
@@ -350,6 +350,52 @@ getVariableName <- function(match_call, varname) {
   idx[!is.na(idx)]
 }
 
+#' find reference run/sample of each group
+#'
+#' @param x exprs matrix
+#' @param group grouping variable, i.e. protein accession
+#' @return column index of reference fraction per group
+#' @noRd
+.referenceFraction <- function(x, group) {
+  notNA <- !is.na(x)
+  mode(notNA) <- "numeric"
+  rs <- rowsum(x, group=group, reorder=FALSE, na.rm=TRUE)
+  rsNA <- rowsum(notNA, group=group, reorder=FALSE, na.rm=TRUE)
+  setNames(max.col((.rowMaxs(rsNA) == rsNA) * rs), rownames(rs))
+}
+
+#' norm to reference run/sample for each protein
+#'
+#' described in: https://doi.org/10.1104/pp.114.245589
+#' The peptide data were converted to protein intensities as follows. For each
+#' protein, a fraction with the highest number of peptides quantified was
+#' nominated as a reference fraction. The protein abundance in that reference
+#' fraction was taken as one. Then, other fractions were quantified against the
+#' reference fraction by finding the ratio of the combined intensity for
+#' peptides shared between the interrogated and reference fractions. When the
+#' quantities in all fractions for a given protein were computed, the values
+#' were renormalized to give a sum = 1 across all 10 fractions.
+#'
+#' @param x exprs matrix
+#' @param group grouping variable, i.e. protein accession
+#' @param norm normalise proteins to sum 1?
+#' @return column index of reference fraction per group
+#' @noRd
+.normToReferenceFraction <- function(x, group, norm=TRUE) {
+  ref <- .referenceFraction(x, group)
+  nr <- nrow(x)
+  mRef <- x
+  mRef[] <- x[(ref[group] - 1L) * nr + 1L:nr]
+  x[is.na(mRef)] <- NA_real_ # mark all values as NA where reference is NA
+  mRef[is.na(x)] <- NA_real_
+  r <- rowsum(x, group=group, reorder=FALSE, na.rm=TRUE) /
+        rowsum(mRef, group=group, reorder=FALSE, na.rm=TRUE)
+  if (norm) {
+    r <- r/rowSums(r)
+  }
+  r
+}
+
 ## Computes header from assay data by-passing cache
 .header <- function(object) {
   if (length(object) == 0)
--- a/R/utils.R
+++ b/R/utils.R
@@ -289,6 +289,22 @@ getVariableName <- function(match_call, varname) {
   tail(as.character(mcx), n = 1)
 }
 
+#' rowwise max, similar to rowwise mean via rowMeans
+#'
+#' @param x matrix
+#' @param na.rm logical
+#' @return double vector with maximum values per row
+#' @seealso Biobase::rowMax (could not handle missing values/NA)
+#' @noRd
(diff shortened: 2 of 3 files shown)


pavel-shliaha:
probably should go into MSnBase, after all there is nothing specific in this normalisation method for organelle proteomics. I do not remember why it went into pRoloc before.

sgibb:
@lgatto sorry this PR was not meant to be merged yet. While the workhorse functions are working (and hidden by `NAMESPACE` so nothing changed for the user) there is not any user-visible API to access them nor any documentation. If you don't have a better name I am going to create an user-visible function `normToReference`.

lgatto:
Ok, go ahead and merge when ready.

sgibb:
@pavel-shliaha 
>  the values were renormalized to give a sum = 1 across all 10 fractions.

That means I divided the intensities by the number of runs (otherwise a single identified run would get the intensity 1 and sample with 10 runs would just get 1/10 intensity for the same amount of peptide). Is that right?

pavel-shliaha:
sorry, I am counfused. What I meant by this phrase is I have a table of peptide intensities, whereby rows Aare peptides and columns are runs (LC-MS/MS acquiositions). Before starting the normalsiation, I make the total sum in every column to be equal 1.

sgibb:
Ok, so we have a matrix with `peptides x runs` (rows x columns) and we do the following:

1. Normalise columns (runs) to sum up to 1, that is `t(t(x)/colSums(x))`.
2. Find reference column (run) for each protein (use the column (run) with the lowest number of `NA` if multiple candidates are available use the one with the highest intensity).
3. For each protein and each run: 
3.1 Find peptides shared by the current run and the reference.
3.2 Sum the shared peptides for current run and reference.
3.3 Divided sum of shared peptides of current run by sum of shared peptides of reference run: `sum(sharedPeptidesCurrent)/sum(sharedPeptidesRef)`
3.4 Now rows are proteins (`proteins x runs`)
4.  Normalise rows (proteins) to sum up to 1, that is `x/rowSums(x)`.

@pavel-shliaha my question is: Should we use `x/rowSums(x)` or something else for step 4 (and 1):

Here a MWE:
```r
x <- matrix(c(3, 2, 2, 2, 1, 1, 1, 1, 1, NA, NA, NA), nrow=3, byrow=TRUE,
            dimnames=list(paste0("Peptide", c("A", "A", "B"), c(1, 2, 1)),
                          paste0("Run", 1:4)))
x
#            Run1 Run2 Run3 Run4
# PeptideA1    3    2    2    2
# PeptideA2    1    1    1    1
# PeptideB1    1   NA   NA   NA

# normalise columns to 1
x <- t(t(x)/colSums(x, na.rm=TRUE))
#            Run1      Run2      Run3      Run4
# PeptideA1  0.6 0.6666667 0.6666667 0.6666667
# PeptideA2  0.2 0.3333333 0.3333333 0.3333333
# PeptideB1  0.2        NA        NA        NA

## reference run is 2 for A and 1 for B

## normalise to reference
## (for Protein A all peptides are shared, for B none)
x <- rowsum(x, group=c(1, 1, 2)) / c(sum(x[1:2,2]), x[3, 1])
rownames(x) <- paste0("Protein", c("A", "B"))
#          Run1 Run2 Run3 Run4
# ProteinA  0.8    1    1    1
# ProteinB  1.0   NA   NA   NA

## normalise proteins to one
x/rowSums(x, na.rm=TRUE)
#               Run1      Run2      Run3      Run4
# ProteinA 0.2105263 0.2631579 0.2631579 0.2631579
# ProteinB 1.0000000        NA        NA        NA
```

Now *ProteinB* is fourth to five times more abundant than *ProteinA* and its peptides were of the same intensity initial. So I have two problems here:
1. Normalisation of the columns (runs) to one penalises the runs  with a lot of peptides and proteins: in the MWE *ProteinA* is more abundant in *Run1* than in all other runs. After normalisation of the columns it is the other way around (just because there are no peptides for other proteins in the other runs).
2. Normalisation of the columns (runs) to one changes the reference run: in the MWE the reference run for *ProteinA* (*PeptideA1* and *PeptideA2*) would be *Run1* but because of the normalisation it is *Run2*.
2. Normalisation of the rows (proteins) to one penalises the proteins that were detected in many runs (and supports the proteins that were detected in just a few runs).

Do I misunderstand the algorithm?

JFTR neither step 1 (normalise runs to one) nor step 4 (normalise proteins to one) was part of @lgatto's implementation in *pRoloc*. And as far as I understand your implementation you just applied step 4 (normalise proteins to one) not step 1.

pavel-shliaha:
@sgibb 

1) normalisation should be done on the total signal for all peptides within a run, NOT for peptides within a protein. This should actually be roughly true anyway, since I should be loading the same amount of digest on a column. Thus if in one run 50% of intensity is contributed by protein A and in another only 25% intensity is protein A, then the ratio of protein amount should be 2:1. Think about this more conceptually. What are we measuring? We are measuring proportion of signal for a particular protein in 1ug of analysed material. NB: any other normalisation (e.g. quantile normalisation?) can be performed on a dataset, so perhaps leave it out of the new function. However for organelle proteomics datasets I would still do normalisation to match total signals between runs

2) you dont have to normalise to 1. Its just as far as I remember pRoloc will complain without this normalisation. I would not normalise to 1 within this function and let user renormalise to 1 if they choose.

bottom line: I would leave steps 1 and 4 out of the final function.

sgibb:
ok, thanks for clarification. So I remove the both normalisation (step 1 and 4). You are right, it could be easily done later.

sgibb:
`combineFeaturures` gains a new `fun` named `NTR` that implements the algorithm discussed above.

Issue #214, opened 2017-05-29
Reading mzData in readMSData2() broken with default mzR:pwiz backend


sneumann:
Hi, 
the new default backend in mzR is pwiz, but mzXML and mzData are only handled 
by the mzR:RAMP backend:

```
> fticr <- readMSData2(list.files(system.file("fticr", package = "msdata"), recursive = TRUE, full.names = TRUE)[1], msLevel. = 1)
Error in pwizModule$open(filename) : 
  [MSDataFile::readFile()] Unsupported file format.
```

In XCMS we check the required backend like this:
https://github.com/sneumann/xcms/blob/ed8945662d82612911bcdbf4be3333ea750d1ef6/R/xcmsSource.R#L9
and use the corresponding backend:
https://github.com/sneumann/xcms/blob/ed8945662d82612911bcdbf4be3333ea750d1ef6/R/xcmsSource.R#L9
Yours, Steffen

jotsetung:
Fixing this results in a log of unit test failures in `MSnbase` - have to check that. Possible reason: if `pwiz` is used as backend stuff like `polarity` is correctly read, while `Ramp` doesn't.

sneumann:
Hi, I can conform that the xcms `R CMD check` passes with
```
other attached packages:
[1] xcms_2.99.2         MSnbase_2.3.4       ProtGenerics_1.6.0 
[4] mzR_2.11.3          Rcpp_0.12.10        BiocParallel_1.8.2 
[7] Biobase_2.34.0      BiocGenerics_0.20.0
```
all pulled tonight from github. Well done everyone ! Yours, Steffen

sneumann:
Hi, I can conform that the xcms `R CMD check` passes with
```
other attached packages:
[1] xcms_2.99.2         MSnbase_2.3.4       ProtGenerics_1.6.0 
[4] mzR_2.11.3          Rcpp_0.12.10        BiocParallel_1.8.2 
[7] Biobase_2.34.0      BiocGenerics_0.20.0
```
all pulled tonight from github. Well done everyone ! Yours, Steffen

PR #217, opened 2017-05-31
Switch mzR backend based on data type

jotsetung:
- Fix for issue #214: switch mzR backend based on the file type.
- Add unit tests for the new mzRBackend function.
- Fix unit tests in test_mzR-funs.R failing after switch to the new backend.

### DIFF
--- a/R/functions-OnDiskMSnExp.R
+++ b/R/functions-OnDiskMSnExp.R
@@ -232,11 +232,8 @@ precursorValue_OnDiskMSnExp <- function(object, column) {
     if (missing(fData) | missing(filenames))
         stop("Both 'fData' and 'filenames' are required!")
     filename <- filenames[fData[1, "fileIdx"]]
-    ## Open the file.
-    if (isCdfFile(filename))
-        fileh <- mzR::openMSfile(filename, backend = "netCDF")
-    else
-        fileh <- mzR::openMSfile(filename)
+    ## issue #214: define backend based on file format.
+    fileh <- mzR::openMSfile(filename, backend = mzRBackend(filename))
     ## hd <- header(fileh)
     on.exit(expr = mzR::close(fileh))
     ## Intermediate #151 fix. Performance-wise would be nice to get rid of this.
--- a/R/functions-OnDiskMSnExp.R
+++ b/R/functions-OnDiskMSnExp.R
@@ -233,7 +233,7 @@ precursorValue_OnDiskMSnExp <- function(object, column) {
         stop("Both 'fData' and 'filenames' are required!")
     filename <- filenames[fData[1, "fileIdx"]]
     ## issue #214: define backend based on file format.
-    fileh <- mzR::openMSfile(filename, backend = mzRBackend(filename))
+    fileh <- .openMSfile(filename)
     ## hd <- header(fileh)
     on.exit(expr = mzR::close(fileh))
     ## Intermediate #151 fix. Performance-wise would be nice to get rid of this.
@@ -369,7 +369,7 @@ precursorValue_OnDiskMSnExp <- function(object, column) {
         stop("Both 'fData' and 'filenames' are required!")
     filename <- filenames[fData[1, "fileIdx"]]
     ## Open the file.
-    fileh <- mzR::openMSfile(filename)
+    fileh <- .openMSfile(filename)
     hd <- header(fileh)
     on.exit(expr = mzR::close(fileh))
     msLevel1 <- which(fData$msLevel == 1)
--- a/R/utils.R
+++ b/R/utils.R
@@ -1068,19 +1068,39 @@ countAndPrint <- function(x) {
 #' @return A \code{character(1)} with the name of the backend (either
 #'     \code{"netCDF"}, \code{"Ramp"} or \code{"pwiz"}.
 #' 
-#' @author Johannes Rainer
+#' @author Johannes Rainer, Sebastian Gibb
 #'
 #' @noRd
-mzRBackend <- function(x) {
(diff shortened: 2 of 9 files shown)


jotsetung:
Just one note - unit tests will run very long until the pull request https://github.com/sneumann/mzR/pull/107 in `mzR` is merged.

R/functions-OnDiskMSnExp.R
```diff
@@ -232,11 +232,8 @@ precursorValue_OnDiskMSnExp <- function(object, column) {
     if (missing(fData) | missing(filenames))
         stop("Both 'fData' and 'filenames' are required!")
     filename <- filenames[fData[1, "fileIdx"]]
-    ## Open the file.
-    if (isCdfFile(filename))
-        fileh <- mzR::openMSfile(filename, backend = "netCDF")
-    else
-        fileh <- mzR::openMSfile(filename)
+    ## issue #214: define backend based on file format.
+    fileh <- mzR::openMSfile(filename, backend = mzRBackend(filename))
```
sgibb: Because you always use the same arguments for `openMSfile` (and the user is not allowed to change these and it is too easy to forget the `backend` argument), I would suggest to create a new function for this:
```r
.openMSfile <- function(filename) {
  mzR::openMSfile(filename, backend=mzRBackend(filename))
}
```

R/utils.R
```diff
@@ -1057,3 +1057,30 @@ countAndPrint <- function(x) {
...
+#'
+#' @param x \code{character(1)} representing the file name.
+#'
+#' @return A \code{character(1)} with the name of the backend (either
+#'     \code{"netCDF"}, \code{"Ramp"} or \code{"pwiz"}.
+#' 
+#' @author Johannes Rainer
+#'
+#' @noRd
+mzRBackend <- function(x) {
```
sgibb: AFAIK `mzR::openMSfile` supports just one filename and just one single backend. That's why something like `mzRBackend(c("foo.mzML", "foo.xml", "foo.nc"))` should fail (and not return `c("pwiz", "netCDF")`). Or it should be completely vectorized and return `c("pwiz", "pwiz", "netCDF")`. But IMHO vectorization is overkill here (because `mzR::openMSfile` expects one single filename and backend).

R/utils.R
```diff
@@ -1057,3 +1057,30 @@ countAndPrint <- function(x) {
...
+    backends <- c(pwiz = "\\.mzml($|\\.)|\\.mzxml($|\\.)",
+                  Ramp = "\\.mzdata($|\\.)",
+                  netCDF = "\\.cdf($|\\.)|\\.nc($|\\.)")
+    matches <- sapply(backends, function(z) grep(z, x, ignore.case = TRUE),
```
sgibb: I am not sure how often we call this function. If we avoid vectorization I would suggest to use:
```r
if (grepl("\\.mzml($|\\.)|\\.mzxml($|\\.)", x)) {
  return("pwiz")
} else if (grepl("\\.mzdata($|\\.)", x)) {
  return("Ramp")
} else if (grepl("\\.cdf($|\\.)|\\.nc($|\\.)", x)) {
  return("netCDF")
} else {
  stop("Could not determine file type for ", x)
}
```
This would avoid some looping, regex evaluation and comparisons in the most used cases (e.g. pwiz).

R/functions-OnDiskMSnExp.R
jotsetung: makes sense. Will do.

R/utils.R
jotsetung: Makes sense, thanks for the suggestion.

R/utils.R
jotsetung: OK

jotsetung:
Thanks @sgibb for the review!  
Let's wait now for @lgatto for the final OK, merge and commit to Bioc.

tests/testthat/test_mzR-funs.R
```diff
@@ -36,31 +36,36 @@ test_that(".chomatogram function", {
...
+    close(ms)
 })
 
 
 test_that("chomatogram methods", {
     f <- dir(system.file("microtofq", package = "msdata"),
              full.names = TRUE, pattern = "MM8.mzML")
     library("mzR")
-    ms <- openMSfile(f)
+    ## Note: this does NOT work yet with backend = "pwiz"!
```
lgatto: Could you open an issue for this, to keep track of this.

R/utils.R
```diff
@@ -1057,3 +1057,50 @@ countAndPrint <- function(x) {
...
+#'
+#' @param x \code{character(1)}: the file name.
+#'
+#' @return A file handle to the opened MS file.
+#'
+#' @author Johannes Rainer
+#' 
+#' @noRd
+.openMSfile <- function(x) {
+    if(missing(x) || length(x) != 1)
```
lgatto: I'm being picky here, but could you add a space:
```r
if (missing(x) || length(x) != 1)
```

tests/testthat/test_mzR-funs.R
lgatto: Could you open an issue for this, to keep track of this.

R/utils.R
jotsetung: of course! missed that somehow

jotsetung: Uh, but you merged it before I pushed this last change...

lgatto: Oups, sorry, I thought it was done :-/. I think you can just commit that change.

jotsetung: OK, pushed it to master now. Thanks for merging!

Issue #215, opened 2017-05-30
Spectrum2 C level constructor fails for empty spectra


jotsetung:
The C-constructor for Spectrum2 classes fails if a spectrum does not contain any values. Data from
https://github.com/crmclean/Reproducible_Examples/tree/master/xcms
```r
library(MSnbase)

raw_data <- readMSData2("mtab_mmetsp_ft_120815_01.mzML")

ms1 <- raw_data[[1]]

ms2 <- raw_data[[2]]

Error in Spectra2_mz_sorted(msLevel = msnfd$msLevel, peaksCount = nValues,  : 
  Arguments 'mz', 'intensity' and 'nvalues' are required!

```

jotsetung:
Note, reading just the MS2 spectra using `readMSData` works:
```r
raw_data2 <- readMSData("mtab_mmetsp_ft_120815_01.mzML")
ms2 <- raw_data2[[1]]
mz(ms2)
numeric(0)
```

lgatto:
`Spetra2_mz_sorted` fails as expected [here](https://github.com/lgatto/MSnbase/blob/master/R/functions-Spectrum2.R#L97-L98)

```r
    if (length(mz) == 0 | length(intensity) == 0 | length(nvalues) == 0) {
        stop("Arguments 'mz', 'intensity' and 'nvalues' are required!")
```

I think it would be good to change this so that when `nvalues` is 0 (which is the case above), empty `mz` and `intensity` are allowed, and return a spectrum that is equivalent that the one with the R constructor and `readMSData`.

jotsetung:
Agree. Will do as soon as I find the time for it.

jotsetung:
Fixed. Just removing the check above worked:
```r
library(MSnbase)
raw_data <- readMSData2("mtab_mmetsp_ft_120815_01.mzML")
ms1 <- raw_data[[1]]
ms2 <- raw_data[[2]]
mz(ms2)
numeric(0)

only_ms2 <- filterMsLevel(raw_data, msLevel = 2)
all_ms2 <- spectra(only_ms2)
all_ms2_2 <- spectra(readMSData("mtab_mmetsp_ft_120815_01.mzML", msLevel. = 2))
expect_equal(unname(all_ms2), unname(all_ms2_2))
## Works
```

jotsetung:
Made pull request https://github.com/lgatto/MSnbase/pull/219

PR #219, opened 2017-06-05
Fix issue #215

jotsetung:
- Remove check for parameters being of length > 0.
- Add unit tests to ensure empty spectra are correctly returned.

### DIFF
--- a/R/functions-Spectrum1.R
+++ b/R/functions-Spectrum1.R
@@ -93,12 +93,15 @@ Spectra1_mz_sorted <- function(peaksCount = NULL, rt = numeric(),
                                intensity = numeric(), fromFile = integer(),
                                centroided = NA, smoothed = NA,
                                polarity = NA_integer_, nvalues = integer()) {
-    if (length(mz) == 0 | length(intensity) == 0 | length(nvalues) == 0) {
-        stop("Arguments 'mz', 'intensity' and 'nvalues' are required!")
-    } else {
-        if (length(mz) != length(intensity))
-            stop("Lengths of 'mz' and 'intensity' do not match!")
-    }
+    ## Fix issue #215: remove check to allow empty spectra
+    ## if (length(mz) == 0 | length(intensity) == 0 | length(nvalues) == 0) {
+    ##     stop("Arguments 'mz', 'intensity' and 'nvalues' are required!")
+    ## }
+    if (sum(nvalues) != length(mz))
+        stop("Length of 'mz' does not match with the number of values per ",
+             "spectrum")
+    if (length(mz) != length(intensity))
+        stop("Lengths of 'mz' and 'intensity' do not match!")
     nvals <- length(nvalues)
     ## Now match all of the lengths to the length of nvalues.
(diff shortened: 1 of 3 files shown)


R/functions-Spectrum1.R
```diff
@@ -93,12 +93,15 @@ Spectra1_mz_sorted <- function(peaksCount = NULL, rt = numeric(),
...
-        stop("Arguments 'mz', 'intensity' and 'nvalues' are required!")
-    } else {
-        if (length(mz) != length(intensity))
-            stop("Lengths of 'mz' and 'intensity' do not match!")
-    }
+    ## Fix issue #215: remove check to allow empty spectra
+    ## if (length(mz) == 0 | length(intensity) == 0 | length(nvalues) == 0) {
+    ##     stop("Arguments 'mz', 'intensity' and 'nvalues' are required!")
+    ## }
+    if (sum(nvalues) != length(mz))
```
lgatto: Is this really `sum(nvlaues)`? Shouldn't it be `nvalues`?

lgatto:
I would also suggest to add tests to `test_Spectrum.R` for example, where I test that an empty spectrum (levels 1 and 2) are valid. 

```r
test_that("Spectrum validity", {
  expect_true(validObject(new("Spectrum1")))
  expect_true(validObject(new("Spectrum2")))
  expect_true(validObject(sp3 <- new("Spectrum2", msLevel = 3L)))
  expect_equal(msLevel(sp3), 3L)
})
```

R/functions-Spectrum1.R
jotsetung: `nvalues` is an `integer` of length equal to the number of spectra, each value representing the number of peaks it contains (and hence how many `mz` and `intensity` values it contains). This is used to *split* the `mz` and `intensity` vectors (that contain the mz and intensity values of all spectra) per spectrum. 
The test is thus OK.

lgatto: ok

tests/testthat/test_Spectrum_Constructor.R
```diff
@@ -126,6 +128,30 @@ test_that("C-level multi-Spectrum1 constructor with M/Z ordering", {
         expect_identical(mz(spectL[[i]]), mzValsList[[i]][idxList[[i]]])
         expect_identical(intensity(spectL[[i]]), intValsList[[i]][idxList[[i]]])
     }
+
+    ## Check empty spectra.
+    ## Spectrum1
+    mzVals <- sort(abs(rnorm(200, mean = 100, sd = 10)))
```
sgibb: Should we use `set.seed` here? Otherwise if a test randomly fails we would never able to reproduce it because we don't know the input values.

jotsetung: I don't believe that these tests will ever fail (fingers crossed), but I'll add the `set.seed` to be on the save side.

lgatto: > I don't believe that these tests will ever fail

LOL!

PR #223, opened 2017-06-20
Remove redundant readMSData and readMSData2 calls from unit tests

jotsetung:
Read common test files only once. Should reduce unit test running time a bit.


lgatto:
In addition, I have used the full file name as pattern to `msdata::proteomics` (see #7fb64e1e8dcb4e197f0aed45c0ab90f8ab08397d), as I added the full `TMT_Erwinia` file to `msdata`. The one we had was only a subset.

Issue #225, opened 2017-06-24
spectra ordering for multiple files


sgibb:
This issue is the same as #68 (PR #69) was for spectra located in a single file but for multiple files. If I read multiple files the order is now: spectrum1.file1, spectrum1.file10, spectrum1.file11, ... , spectrum1. file2, ... , spectrum2.file1, ... (`X001.1`, `X001.10`, `X001.11`, ... `X001.2`, ... `X002.1`).

Wouldn't it be more useful to swap the order: file.spectrum and apply natural sorting to files as well?

So, e.g. file1.spectrum1, file1.spectrum2, ..., file2.spectrum1, ... (`X01.001`, `X01.002`, ... `X02.001`).

from `readMSData2.R`:
```r
fullhdorder <- c(fullhdorder,
                     sprintf(paste0("X%0",
                                    ceiling(log10(length(spidx) + 1L)),
                                    "d.%s"), 1:length(spidx), filen))
```
should be:
```r
nDigitsSpIdx <- ceiling(log10(length(spidx) + 1L))
nDigitsFileIdx <- ceiling(log10(length(files) + 1L))

fullhdorder <- c(fullhdorder,
                     sprintf(paste0("X%0", nDigitsFileIdx, "d.",
                                     "%0", nDigitsSpIdx, "d"), 
                                     filen, seq_along(spidx))
```

BTW: I am not sure how this would effect existing code/objects.

CC: @pavel-shliaha

pavel-shliaha:
according to Laurent it is impossible to reorder spectra in MSnBase object. This is a huge inconvinience, since it means that the table with frgmentation parameters needs to be reordered instead

lgatto:
> according to Laurent it is impossible to reorder spectra in MSnBase object. This is a huge inconvinience, since it means that the table with frgmentation parameters needs to be reordered instead

There is no fixed order of spectra in an `MSnExp` or an `OnDiskMSnExp`. You can define the order that suites you best.

lgatto:
I am currently testing, and have found at least one example that needs updating. 

@sgibb, could you elaborate on the need for a specific order, and how, this particular naming scheme is needed in addition to `filterFile`. I'm fine with renaming the features as you suggest, but would appreciate a bit more discussion to see beyond the case where a very fast `filterFile` would work too.

sgibb:
In fact there is no important reason but convenience. (I am afraid I have not any strong argument.) 

[...]

PR #226, opened 2017-06-25
introduce new spectra naming scheme

sgibb:
See discussion in #225.

The new naming schema would be FFILEID.SSPECTRUMID, e.g. `F01.S0001`.

Currently I am at the train and don't have all data files at hand (and my bandwidth is too limited). That's why I can't run the test suite, yet. Sorry.


lgatto:
I will have a go and test it. It would also be useful to get @jotsetung's input when he's back from holiday.

jotsetung:
I like the new naming scheme.

Issue #236, opened 2017-08-01
Add phenoData to Chromatograms class


jotsetung:
What about adding a `pdata` slot to the `Chromatograms` object?
If we add that:
+ Ensure that subsetting also correctly sub-sets the pheno data.
+ Implement all getters and setters (i.e. `$`, `$<-`, `pData`, `pData<-`.

lgatto:
Yes, excellent idea. Will there be an explicit relation between the dimensions of the `Chromatograms` and `phenoData` objects?

jotsetung:
Yup, `nrow(phenoData)` has to be the same than `ncol(Chromatograms)`. I'll ensure that for all subsetting operations etc.

jotsetung:
Made pull request https://github.com/lgatto/MSnbase/pull/239

PR #239, opened 2017-08-02
Add phenoData slot to Chromatograms (issue #236)

jotsetung:
- Add phenoData slot to Chromatograms.
- chromatogram,MSnExp passes the phenoData along to Chromatograms.
- Add methods phenoData, pData, pData<-, $, $<- and colnames<- for Chromatograms
  objects.
- Add related documentation and unit tests.
- [,Chromatograms subsets also the phenoData.


R/methods-Chromatograms.R
```diff
@@ -190,3 +212,52 @@ setMethod("plot", signature = signature("Chromatograms"),
...
+#'     samples. Returns a \code{data.frame}.
+setMethod("pData", "Chromatograms", function(object) pData(phenoData(object)))
+
+#' @rdname Chromatograms-class
+#'
+#' @description \code{pData<-}: replace the phenotype data.
+setReplaceMethod("pData", "Chromatograms", function(object, value) {
+    pData(object@phenoData) <- value
+    if (any(rownames(value) != as.character(1:nrow(value))))
+        colnames(object@.Data) <- rownames(value)
```
lgatto: Why do the rownames of the new `pData` take precedence on the colnames of the current object? If they don't match, I think it would be safer to stop with an error, rather than assuming that `pData`'s rownames are the correct one's.

R/functions-Chromatograms.R
```diff
@@ -20,14 +33,37 @@
...
+    if (ncol(datmat) != nrow(phenoData))
+        stop("Dimensions of the data matrix and the  phenoData do not match")
+    ## Convert phenoData...
+    if (is(phenoData, "data.frame"))
+        phenoData <- AnnotatedDataFrame(phenoData)
+    if (is(phenoData, "AnnotatedDataFrame"))
+        phenoData <- as(phenoData, "NAnnotatedDataFrame")
+    ## Set colnames if we have some
+    if (any(rownames(phenoData) != as.character(1:nrow(phenoData))))
+        colnames(datmat) <- rownames(phenoData)
```
sgibb: Same problem here as @lgatto already highlighted for `pData<-`.

R/methods-Chromatograms.R
```diff
@@ -80,10 +89,18 @@ setMethod("[", "Chromatograms",
...
               ## Multiple elements, return type depends on drop.
               x <- x@.Data[i = i, j = j, drop = drop]
-              if (!drop)
+              if (!drop) {
                   x <- as(x, "Chromatograms")
+                  pd <- pd[j, ]
+                  ## Drop levels
+                  pData(pd) <- droplevels(pData(pd))
+                  ## set row names
+                  rownames(pd) <- colnames(x)
```
sgibb: It is a bit confusing for me. Sometimes the `phenoData` set the `colnames` of `.Data` and here it is the other way around. Which one is more important? Is it really necessary to overwrite this information again and again?

R/methods-Chromatograms.R
```diff
@@ -190,3 +212,52 @@ setMethod("plot", signature = signature("Chromatograms"),
...
+})
+
+#' @rdname Chromatograms-class
+#'
+#' @description \code{$} and \code{$<-}: get or replace individual columns of
+#'     the object's pheno data.
+#'
+#' @param name For \code{$}, the name of the pheno data column.
+setMethod("$", "Chromatograms", function(x, name) {
+    eval(substitute(pData(x)$NAME_ARG, list(NAME_ARG = name)))
```
sgibb: Why we could not simply use `pData(x)[[name]]` here (as in the replace method)?

R/methods-Chromatograms.R
jotsetung: So you mean I should add a check that ensures that, if `object@.Data` has `colnames` that they match the `rownames` of the new `data.frame`? Makes sense.

R/methods-Chromatograms.R
jotsetung: I'll try that.

R/methods-Chromatograms.R
jotsetung: It is required because:
```r
df <- data.frame(a = letters[1:4], idx = 1:4)
## rownames are not NULL
rownames(df)
[1] "1" "2" "3" "4"

## When I sub-set they keep the original row names.
df_sub <- df[2:3, ]
rownames(df_sub)
[1] "2" "3"

## That's no problem if we have *real* rownames set (the 1:nrow(df) above is the 
## case were we don't have rownames specified.
## The problem comes if we don't have rownames/colnames (e.g. a `Chromatogram` 
## without a `phenoData` specified). In that case, subsetting the matrix will 
## have e.g.
mat <- matrix(4, ncol = 4, nrow = 3)
mat[, 2:3]
     [,1] [,2]
[1,]    4    4
[2,]    4    4
[3,]    4    4

## And subsetting the corresponding `data.frame`/phenoData:
df[2:3, ]
  a idx
2 b   2
3 c   3
## With
rownames(df[2:3, ])
[1] "2" "3"

## See, I find that pretty misleading. In the case we don't have rownames/colnames 
## available the subsetted `data.frame` says we have rows with index 2 and 3, and
## the `matrix` says 1 and 2.
## Setting rownames(pd) <- colnames(x) does:
df_sub <- df[2:3, ]
rownames(df_sub) <- NULL
rownames(df_sub)
[1] "1" "2"

## So, for the case in which we do have colnames and rownames defined the assignment
## would not  be necessary, but when `is.null(colnames(x))` it is required.
```
Does that make sense?

R/functions-Chromatograms.R
jotsetung: Well, I didn't think of the possibility of the user passing `dimnames` with `...`. In that case, yes, I have to check that `colnames(datmat)` and `rownames(phenoData)` match. Otherwise, if `if(is.null(colnames(datmat))` I'll have to set `colnames(datmat)` to `rownames(phenoData`, **but only** if `rownames(phenoData)` is not equal to `as.character(1:nrow(phenoData))`.

R/methods-Chromatograms.R
jotsetung: Syncing the `colnames` of `@.Data` and `rownames` of `@phenoData` even if `rownames` of the phenoData are not set would solve also @sgibb 's comment above. I'll implement and check if other problems arise.

R/methods-Chromatograms.R
```diff
@@ -190,3 +212,52 @@ setMethod("plot", signature = signature("Chromatograms"),
...
+})
+#' @rdname Chromatograms-class
+setReplaceMethod("$", "Chromatograms", function(x, name, value) {
+    pData(x)[[name]] <- value
```
jotsetung: In principle `pData(x)[[name]] <- ` throws an error if the lengths do not match, unless `value` is recycled. I did not check to mimic the *standard* R-behavior in such cases.

R/methods-Chromatograms.R
jotsetung: `colnames(object@.Data)` are now always synced with `rownames(object@phenoData)` (commit https://github.com/lgatto/MSnbase/commit/b658d43162c0a964a17709aa8acf64c2c7497d00), so no need anymore for these lines.

R/functions-Chromatograms.R
jotsetung: Removed now in https://github.com/lgatto/MSnbase/commit/b658d43162c0a964a17709aa8acf64c2c7497d00 as no longer required.

R/methods-Chromatograms.R
jotsetung: Fixed in https://github.com/lgatto/MSnbase/commit/b658d43162c0a964a17709aa8acf64c2c7497d00 . Setting the `colnames(object@.Data)` from the beginning to `rownames(phenoData)` fixes this.

R/methods-Chromatograms.R
jotsetung: Done.

R/methods-Chromatograms.R
lgatto: Now I realise that there is this issue in `MSnSet`s too
```r
> data(msnset)
> sampleNames(msnset)
[1] "iTRAQ4.114" "iTRAQ4.115" "iTRAQ4.116" "iTRAQ4.117"
> pData(msnset)
              mz reporters
iTRAQ4.114 114.1    iTRAQ4
iTRAQ4.115 115.1    iTRAQ4
iTRAQ4.116 116.1    iTRAQ4
iTRAQ4.117 117.1    iTRAQ4
> pd <- data.frame(x = 1:4, y = 4:1)
> rownames(pd) <- LETTERS[1:4]
> pd
  x y
A 1 4
B 2 3
C 3 2
D 4 1
> pData(msnset) <- pd
> sampleNames(msnset)
[1] "A" "B" "C" "D"
> validObject(msnset)
Error in validObject(msnset) : 
  invalid class “MSnSet” object: 1: sampleNames differ between assayData and phenoData
invalid class “MSnSet” object: 2: sampleNames differ between phenoData and protocolData
```
I think that the error should be thrown when trying to assign a new `pData` with different rownames.

jotsetung: I thought I did fix that for `Chromatograms`. Will check

jotsetung: OK, it was fixed:
```r
    ints <- abs(rnorm(12, sd = 20))
    ch <- Chromatogram(rtime = 1:length(ints), ints)
    ints <- abs(rnorm(20, sd = 14))
    ch1 <- Chromatogram(rtime = 1:length(ints), ints)
    ints <- abs(rnorm(14, sd = 24))
    ch2 <- Chromatogram(rtime = 1:length(ints), ints)
    ints <- abs(rnorm(40, sd = 34))
    ch3 <- Chromatogram(rtime = 1:length(ints), ints)
    chrs <- Chromatograms(list(ch, ch1, ch2, ch3), nrow = 2)
    ## Check error when assigning a phenoData with different names
    pd <- data.frame(name = letters[1:2], idx = 1:2)
    rownames(pd) <- letters[1:2]
    pData(chrs) <- pd
Error in validObject(object) : 
  invalid class “Chromatograms” object: colnames of object has to match rownames of phenoData
```
I added now also a corresponding unit test.

Issue #250, opened 2017-09-06
Implement a writeMSData,OnDiskMSnExp method


jotsetung:
Implement a `writeMSData,OnDiskMSnExp` method to write the data of an `OnDiskMSnExp` to mzML/mzXML file(s).

stanstrup:
You're a mind reader ;)

stanstrup:
Hi,

I tried to use this today to filter some data since matchedfilter doesn't have the noise parameter centwave does. 
But something seems to go wrong.

I tried this:

```r
library(MSnbase)

raw <- readMSData("160615_0301.CDF", mode="onDisk", msLevel. = 1, centroided. = TRUE)

# polarity(raw) <- 1L # does not work. Should it?
fData(raw)$polarity <- 1L

raw_clean <- removePeaks(raw, 40)
  
writeMSData(raw_clean, "160615_0301_threshold.mzML", outformat = c("mzml"), copy=FALSE)
```

Which gives this error:

```
Error in .local(object, file, ...) : 
  Error checking parameter 'header': 'x' is missing one or more required columns: seqNum, acquisitionNum, msLevel, polarity, peaksCount, totIonCurrent, retentionTime, basePeakMZ, basePeakIntensity, collisionEnergy, ionisationEnergy, lowMZ, highMZ, precursorScanNum, precursorMZ, precursorCharge, precursorIntensity, mergedScan, mergedResultScanNum, mergedResultStartScanNum, mergedResultEndScanNum, injectionTime
```

Seems it is "peaksCount" and "lowMZ" that are missing. I assume because you cannot get them from CDF without reading the whole file?

I therefore tried the inMemory method but after ½ h I gave up. It seems to be taking forever to read the data into memory with this file.

jotsetung:
Thanks for reporting! I'll check

jotsetung:
Can you check the `colnames(fData(raw))` and `colnames(fData(raw_clean))` please? what are they?

stanstrup:
If I don't do anything (apart from the removePeaks of course) I get

```r
> colnames(fData(raw))
 [1] "fileIdx"                  "spIdx"                    "centroided"               "smoothed"                
 [5] "seqNum"                   "acquisitionNum"           "msLevel"                  "originalPeaksCount"      
 [9] "totIonCurrent"            "retentionTime"            "basePeakMZ"               "basePeakIntensity"       
[...]

PR #253, opened 2017-09-18
Add write,MSnExp and write,OnDiskMSnExp

jotsetung:
- Add `write,MSnExp` and `write,OnDiskMSnExp` to allow saving these objects to mzML
  or mzXML file(s) (issue #250).
- Add related documentation and unit tests.

Naming of the methods is open for discussion; I've called them `write` because in `mzR` there is already a `writeMSData` function. Reasoning for calling the function just `write`: we do have MS data stored in the object, calling `write` on them should be enough for the user to know what happens - `writeMSData` basically means *write the MS data in the provided object that contains MS data`, `write`: *write the provided object that contains MS data`.

Note also that the unit tests fail without the latest fix (version 2.11.10) in `mzR`.


R/writeMSData.R
```diff
@@ -0,0 +1,186 @@
...
+        } else {
+            ## MSnExp: feature data does not provide all the data we need.
+            hdr <- data.frame(do.call(rbind,
+                                      spectrapply(z, FUN = .spectrum_header,
+                                                  BPPARAM = SerialParam())),
+                              check.names = FALSE)
+        }
+        ## o Re-calculate stuff we don't have or which might have been
+        ##   changed:
+        new_vals <- do.call(rbind, lapply(pks, function(sp) {
```
sgibb: You are creating exact these values in `.spectrum_header` already. How could they change from line 64 to line 71 without being manipulated by us? If only the `fData` values could have been modified this `do.call` and the following `hdr$...` should go into the `if(is(x, "OnDiskMSnExp"))` block above (line 61).

R/writeMSData.R
```diff
@@ -0,0 +1,186 @@
...
+        baseline = "MS:1000543",
+        alignment = "MS:1000745"
+    )
+    ## Note: we are matching each names(.PATTERN.TO.CV) to `pattern`, not the
+    ## other way round. So we can provide a string describing the processing
+    ## step and return the best matching CV terms.
+    matches <- vapply(names(.PATTERN.TO.CV), FUN = function(z) {
+        length(grep(pattern = z, x = pattern, ignore.case = TRUE)) > 0
+    }, FUN.VALUE = logical(1))
+    if (any(matches))
```
sgibb: Instead of "preallocating" `res` in the first line. We could simply use
```r
if (any(matches))
    .PATTERN.TO.CV[matches]
else
   ifnotfound
```

R/writeMSData.R
```diff
@@ -0,0 +1,186 @@
...
+        centroid = "MS:1000035",
+        smooth = "MS:1000542",
+        baseline = "MS:1000543",
+        alignment = "MS:1000745"
+    )
+    ## Note: we are matching each names(.PATTERN.TO.CV) to `pattern`, not the
+    ## other way round. So we can provide a string describing the processing
+    ## step and return the best matching CV terms.
+    matches <- vapply(names(.PATTERN.TO.CV), FUN = function(z) {
+        length(grep(pattern = z, x = pattern, ignore.case = TRUE)) > 0
```
sgibb: Isn't `any(grepl(...))` easier to read?

R/options.R
```diff
@@ -60,5 +60,9 @@ setMSnbaseFastLoad <- function(opt = TRUE) {
 
 ##' @rdname MSnbaseOptions
 isMSnbaseFastLoad <- function() {
-    MSnbaseOptions()$fastLoad
+    fast_load <- MSnbaseOptions()$fastLoad
+    ## For some odd reasons we get also NULL back - parallel processing?
+    if (!length(fast_load))
+        fast_load <- FALSE
+    fast_load
 }
```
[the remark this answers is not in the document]
jotsetung: I also thought so. But I got very strange errors that took me one day to hunt down: seems that sometimes `MSnbaseOptions()$fastload` is simply not set and `NULL` is returned. This was somehow related to parallel processing.

Re: `camelCase`, I know, I'm inconsistent here. I used to use always only camelCase, but now I got used to snake case because I find it reads better - I can change back if you don't like.

R/writeMSData.R
```diff
@@ -0,0 +1,186 @@
...
+        stop("'files' is required!")
+    if (length(files) != length(fileNames(x)))
+        stop("length of 'files' has to match the number of samples")
+    if (verbose)
+        message("Writing ", length(files), " ", outformat, " files.")
+    ## Split per file.
+    x_split <- splitByFile(x, f = factor(fileNames(x)))
+    ## Using mapply below - in principle we could then even switch to
+    ## bpmapply to perform parallel saving of files.
+    res <- mapply(x_split, files, FUN = function(z, f, outformat,
```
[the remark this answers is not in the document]
jotsetung: Makes sense. OK, I'll do.

R/writeMSData.R
jotsetung: You see everything ;) For `MSnExp` that's correct, for `OnDiskMSnExp` not, because the `fData` for these objects is in fact the `header` at the time when the file is read. So it's better/saver to (re)calculate for `OnDiskMSnExp`.
Yes, I'll put that into the `if(is...`. :+1:

R/writeMSData.R
jotsetung: :+1:

R/writeMSData.R
jotsetung: I'll do.

lgatto:
Regarding `write` vs. `writeMSData`: I surely like `write`, although the latter matches nicely `readMSData`. My only suggestion with `write` would be to ask on the bioc-devel list if they don't want to add the `write` generic to `BiocGenerics`.

lgatto:
See an apparent bug in #255.

R/writeMSData.R
```diff
@@ -44,58 +44,68 @@
...
-            stop("File ", f, " does already exist! Please use a different ",
-                 "file name.")
-        ## o get all peak data. Calling spectrapply will apply also all lazy
-        ##   processing steps. We're not calling the `as` method to avoid R
-        ##   having to look through all namespaces to find the function.
-        pks <- spectrapply(z, FUN = function(sp) cbind(sp@mz, sp@intensity),
-                           BPPARAM = SerialParam())
-        if (is(x, "OnDiskMSnExp")) {
-            hdr <- fData(z)
+    res <- mapply(FUN = .writeSingleMSData, x_split, files,
```
sgibb: The assignment to `res` is not needed here. Just `mapply(...) }`, so the return value of `mapply` will be the return value of `.writeMSData`.

R/writeMSData.R
```diff
@@ -0,0 +1,196 @@
...
+                       BPPARAM = SerialParam())
+        if (is(msData, "OnDiskMSnExp")) {
+            hdr <- fData(msData)
+            ## o Re-calculate stuff we don't have or which might have been
+            ##   changed:
+            new_vals <- do.call(rbind, lapply(pks, function(sp) {
+                cbind(peaksCount = nrow(sp),
+                      totIonCurrent = sum(sp[, 2], na.rm = TRUE),
+                      basePeakMZ = sp[base::which.max(sp[, 2]), 1][1],
+                      basePeakIntensity = max(sp[, 2], na.rm = TRUE))
```
sgibb: Why you calculating the maximum twice? We could simply just call `base::which.max` once and get nearly 20% speed improvement (I guess this is a noncritical function but anyway):
```r
library("msdata")
library("MSnbase")
library("rbenchmark")

m <- readMSData(msdata::proteomics("Top", full.names=TRUE), mode="onDisk")

pks <- spectrapply(m, function(sp)cbind(sp@mz, sp@intensity))

f1 <- function(pks) {
    do.call(rbind, lapply(pks, function(sp) {
        cbind(peaksCount = nrow(sp),
              totIonCurrent = sum(sp[, 2], na.rm = TRUE),
              basePeakMZ = sp[base::which.max(sp[, 2]), 1][1],
              basePeakIntensity = max(sp[, 2], na.rm = TRUE))
}))}

f2 <- function(pks) {
    do.call(rbind, lapply(pks, function(sp) {
        i <- base::which.max(sp[,2])[1]
        cbind(peaksCount = nrow(sp),
              totIonCurrent = sum(sp[, 2], na.rm = TRUE),
              basePeakMZ = sp[i, 1],
              basePeakIntensity = sp[i, 2])
}))}

benchmark(f1(pks), f2(pks), order="relative")
#      test replications elapsed relative user.self sys.self user.child sys.child
# 2 f2(pks)          100   7.791    1.000     7.764    0.004          0         0
# 1 f1(pks)          100   9.447    1.213     9.360    0.008          0         0

all.equal(f1(pks), f2(pks))
# [1] TRUE
```

R/writeMSData.R
jotsetung: I don't want to return anything - so I'll use `invisible` instead.

R/writeMSData.R
jotsetung: good point.

R/writeMSData.R
sgibb: Ok, I thought `.writeSingleMSData` (or `mzR::writeMSData`) would return a value for success/error. But if not `invisible` is fine.

jotsetung:
For the time being I would skip the implementation of the `merge = TRUE` (i.e. merging all data into a single mzML file), since it looks to be a little more complicated and I don't have that much time presently. That OK for you @lgatto and @sgibb ?

Btw, since I changed `write` to `writeMSData` this will require my pull request in `mzR` (https://github.com/sneumann/mzR/pull/133) being merged.

lgatto:
- `mzR` PR merged and pushed to bioc.
- Ok, let's start with multiple file write support, but let's keep it high on the priority list. 

I still need to do some more reviewing, but I haven't had the time so far.

R/writeMSData.R
```diff
@@ -75,17 +75,23 @@
...
+            ##   files (issue #250).
             new_vals <- do.call(rbind, lapply(pks, function(sp) {
                 max_pos <- base::which.max(sp[, 2])[1]
                 cbind(peaksCount = nrow(sp),
                       totIonCurrent = sum(sp[, 2], na.rm = TRUE),
                       basePeakMZ = sp[max_pos, 1][1],
-                      basePeakIntensity = sp[max_pos, 2])
+                      basePeakIntensity = sp[max_pos, 2],
+                      lowMZ = min(sp[, 1]),
+                      highMZ = max(sp[, 1]))
```
sgibb: Isn't `lowMZ = sp[1, 1]` and `highMZ = sp[nrow(sp), 1]`? Or can't we assume correct order at this stage? (I am not sure anymore but if I remember correctly you ensure ordering of mz values in the ctor for spectra.)

jotsetung: You are right - that simplifies it (assuming correct ordering of m/z, but if I'm not mistaken that is also checked in the validity method of `Spectrum` objects).

lgatto:
A few more comments/suggestions/observations;

## The test

```r
library("MSnbase")
f <- msdata::proteomics(full.names = TRUE)[3]
## in memory, MS2 only
x0 <- readMSData(f)
writeMSData(x0, file = "write.mzML")
x1 <- readMSData("write.mzML")
## on disk, MS1 and MS2
x2 <- readMSData(f, mode = "onDisk")
writeMSData(x2, file = "write2.mzML")
x3 <- readMSData("write2.mzML", mode = "onDisk")
```

## File exists

When writing to a file that exists, the error 
```r
> x0 <- readMSData(f)
> writeMSData(x0, file = "write.mzML")
Error in (function (msData, file, outformat, software_processing = NULL,  : 
  File write.mzML does already exist! Please use a different file name.
```
should happen immediately. With my test (see below), it takes 10 seconds. There should be something like 
```r
if (file.exists(file))
  stop("File ", file, " does already exist! Please use a different file name.")
```
right at the beginning of the `writeMSData` method.

## File size

This is not an issue as such, and the difference is relatively small, but it would be useful to know where the difference in file size stems from, and be explicit in the man page in what isn't retained when reading and writing.
```r
> file.info(f)[[1]]
[1] 314974121
> file.info("write2.mzML")[[1]]
[1] 437995651
> file.info(f)[[1]] / file.info("write2.mzML")[[1]]
[1] 0.7191261
```
## Differences 

```r
> all.equal(x2, x3)
Note: method with signature ‘OnDiskMSnExp#MSnExp’ chosen for function ‘all.equal’,
 target signature ‘OnDiskMSnExp#OnDiskMSnExp’.
 "MSnExp#OnDiskMSnExp" would also be valid
[1] "Spectra at position 1 are different."
[2] "Spectra at position 1 are different."
> all.equal(x2[[1]], x3[[1]])
[1] "Attributes: < Component “tic”: Mean relative difference: 2.343212 >"
> tic(x2[[1]])
[1] 9187505
> tic(x3[[1]])
[1] 30715775
```

For the in-memory `MSnExp` objects that contain MS2 data only:
```r
> options(max.print = 10)
> all.equal(x0, x1)
 [1] "Attributes: < Component “assayData”: Component “F1.S0001”: Attributes: < Component “precScanNum”: Mean relative difference: 1 > >"      
 [2] "Attributes: < Component “assayData”: Component “F1.S0001”: Attributes: < Component “scanIndex”: Mean relative difference: 0.9955947 > >"
 [3] "Attributes: < Component “assayData”: Component “F1.S0002”: Attributes: < Component “precScanNum”: Mean relative difference: 1 > >"      
 [4] "Attributes: < Component “assayData”: Component “F1.S0002”: Attributes: < Component “scanIndex”: Mean relative difference: 0.9912281 > >"
 [5] "Attributes: < Component “assayData”: Component “F1.S0003”: Attributes: < Component “precScanNum”: Mean relative difference: 1 > >"      
 [6] "Attributes: < Component “assayData”: Component “F1.S0003”: Attributes: < Component “scanIndex”: Mean relative difference: 0.9868996 > >"
 [7] "Attributes: < Component “assayData”: Component “F1.S0004”: Attributes: < Component “precScanNum”: Mean relative difference: 1 > >"      
 [8] "Attributes: < Component “assayData”: Component “F1.S0004”: Attributes: < Component “scanIndex”: Mean relative difference: 0.9826087 > >"
 [9] "Attributes: < Component “assayData”: Component “F1.S0005”: Attributes: < Component “precScanNum”: Mean relative difference: 1 > >"      
[10] "Attributes: < Component “assayData”: Component “F1.S0005”: Attributes: < Component “scanIndex”: Mean relative difference: 0.978355 > >" 
 [ reached getOption("max.print") -- omitted 12206 entries ]
```
and 
```r
> head(precScanNum(x0))
F1.S0001 F1.S0002 F1.S0003 F1.S0004 F1.S0005 F1.S0006 
     226      226      226      226      226      226 
> head(precScanNum(x1))
F1.S0001 F1.S0002 F1.S0003 F1.S0004 F1.S0005 F1.S0006 
       0        0        0        0        0        0 
> head(scanIndex(x0))
F1.S0001 F1.S0002 F1.S0003 F1.S0004 F1.S0005 F1.S0006 
     227      228      229      230      231      232 
> head(scanIndex(x1))
F1.S0001 F1.S0002 F1.S0003 F1.S0004 F1.S0005 F1.S0006 
       1        2        3        4        5        6 
```
I believe the `scanIndex` difference relates to the discussion we had in #257: it doesn't start with 1 in `x0` because only MS2 spectra are loaded, and when writing on disc, the index always starts at 1. But I think the precusor scan numbers should ideally be preserved.

jotsetung:
Regarding file size: one reason for the difference in file size is that the original file `f` is a gzipped mzML file, while the exported mzML file is uncompressed mzML with only the binary data zlib compressed.

Apart from that, what is not copied/saved is all chromatogram information in the original file. In addition there will be differences in the processing information (i.e. depending on `copy` run-specific metadata is copied over from the original file, such as instrument info, processing steps etc).
I added some more information to the details section of the `writeMSData,MSnExp` method.

jotsetung:
Regarding the differences: (as also stated in the note section of `writeMSData,MSnExp`) total ion current, peak count, base peak m/z and base peak intensity are calculated from the actual spectrum data prior to saving them. Thus you see the differences there in the TIC. Seems the original file contained the ones from the machine.

Regarding the `precScanNum`: that's a tricky situation. I think keeping the `precursorScanNum` is problematic, because it might lead to linking to a wrong spectrum. For the example above it would link to the spectrum number 226, which in the new mzML file is however a MS2 spectrum. That's why I decided to set the `precursorScanNum` to 0. Note that `precursorMz`, `precursorIntensity` and `precursorCharge` are retained. Also, `proteowizard` throws an error if the precursor with the specified ID (scanNum) is not available in the `MSData`.

I did update the `writeMSData,MSnExp` help page and added a sentence explaining this in the *note* section.

lgatto:
@jotsetung - have you added a mention of `writeMSData` in the *demo* and *io* vignettes?

jotsetung:
No I didn't, but will.

jotsetung:
I added a small section to both vignettes and pushed them to the `writeMSData` branch (didn't want to push directly to master). If you're happy with it you can merge it into master.

lgatto:
If you are familiar with `tikz`, you could also update `MSnbase-io-out.tex` to add an arrow from `MSnExp` to raw data. If now, I'll do it tonight.

jotsetung:
I'd prefer if you could do that - I'm currently busy fixing other stuff too...

lgatto:
On Bioc now

Issue #275, opened 2017-11-17
Reading files depends on filename extension


sneumann:
Hi, determining the file type depends on the filename extension. 
In galaxy workflows, input data is unfortunately passed as "something.dat":
```
> MS2RawFile<-readMSData("MM8.mzML", msLevel=1)
> MS2RawFile<-readMSData("MM8.dat", msLevel=1)
Error in .mzRBackend(x) : Could not determine file type for MM8.dat
```
We had the same issue in xcms: 
https://github.com/sneumann/xcms/issues/188 
and @jotsetung implemented a fix in https://github.com/sneumann/xcms/commit/7ef70598754cffc9e896ddb3358f62471e0d16cb

It might even make sense to move the implementation 
of `mzRBackendFromFilename` etc. into mzR itself.

Yours, Steffen

jotsetung:
The trick was to read the first few lines of input if the extension was not recognized and look for some keywords there. I can make a pull request for that.

sgibb:
IMHO @sneumann is right. The whole `mzRBackendFrom*` stuff should be part of `mzR`.

lgatto:
Yes, fine by me to move it to `mzR`; it does make sense indeed.

jotsetung:
Reopening because support is not yet in `MSnbase`

PR #276, opened 2017-11-17
Guess file type from file content (issue #275)

jotsetung:
- If the file extension is unknown, guess file type by its content.
- Works also for compressed files.
- Add related unit tests.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -1161,8 +1161,29 @@ countAndPrint <- function(x) {
     } else if (grepl("\\.cdf($|\\.)|\\.nc($|\\.)", x, ignore.case = TRUE)) {
         return("netCDF")
     } else {
-        stop("Could not determine file type for ", x)
+        return(.mzRBackendFromContent(x))
     }
+}
+
+#' Determine the backend from the (first few lines of the) file content.
+#' 
+#' @author Johannes Rainer
+#'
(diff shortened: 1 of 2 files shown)


R/utils.R
```diff
@@ -1161,10 +1161,31 @@ countAndPrint <- function(x) {
...
+#' @author Johannes Rainer
+#'
+#' @noRd
+.mzRBackendFromContent <- function(x = character()) {
+    if (length(x) != 1)
+        stop("parameter 'x' has to be of length 1")
+    suppressWarnings(
+        first_lines <- readLines(x, n = 4)
+    )
+    if (any(grepl("<mzML", first_lines)) | any(grepl("<mzXML", first_lines))) {
```
sgibb: I would suggest `any(grepl("<mz[X]?ML", first_lines)` here to avoid two `grepl` calls. 

BTW: if two `grepl` calls are needed I would also suggest `||` to avoid the second call if the first is already `TRUE`.

jotsetung:
I'll make a pull request in `mzR` - let's implement this bottom up.

jotsetung:
OK, pull request is in in `mzR` (https://github.com/sneumann/mzR/pull/139). Once that is merged I'll adapt `MSnbase` accordingly.

PR #279, opened 2017-11-25
Link issue numbers to github issues in NEWS.md

sgibb:
IMHO it would be great to look at the `NEWS.md` and just click on an issue number to open the corresponding issue. Unfortunately github doesn't support autolinking inside markdown documents. So I turned all issue numbers into absolute links, please find a preview here:
https://github.com/lgatto/MSnbase/blob/fae29d74b99b2218846d316dadebfbe892d70c06/NEWS.md

It would be a little bit cumbersome to type always type `[#XXX](https://github.com/lgatto/issues/XXX)` into the `NEWS.md` but this effort has to be done once for each issue (and could be automatted see PS2) and saves time searching the issues on github (or copy&paste the issue number).

What do you think?

PS: The `maker` target `NEWS` could be easily modified to remove these links again. So `make PKG=MSnbase NEWS` will produce the current `NEWS` file.
PS2: We could add a `pre-commit` hook to avoid typing these URLs:
```bash
# Replace #XXX issue numbers by [#XXX](https://github.com/lgatto/MSnbase/issues/XXX
sed -i '/\(^\|[^[]\)#[0-9]\+[^]]\?/{s/#\([0-9]\+\)/[#\1](https:\/\/github.com\/lgatto\/MSnbase\/issues\/\1)/g;h};${x;/./{x;q1};x;q0}' NEWS.md
## explanation:
## sed -i (in place replacement)
## /[^[]#[0-9]\+[^]]/ search for issues #XXX not already surounded by "["/"]"
## and apply the substition command just to this numbers
## {s/#\([0-9]\+\)/[#\1](https:\/\/github.com\/lgatto\/MSnbase\/issues\/\1)/g;}
## build the link
## "h" copy the result into the "hold space" (initial empty)
## ${x;/./{x;q1};x;q0} when add the end of file "$", swap "hold" and "pattern"
## space via "x", if now anything was found in the "pattern" space "/./" swap
## again and return 1 (q1); if nothing was found also swap and return 0 (q0)

if [[ $? -ne 0 ]]; then
  echo "Issue link in NEWS.md was modified by 'pre-commit' hook; please review and recommit"
  exit 1
fi
```
PS3: While we are on it: we could create `NEWS` from `NEWS.md` in the `pre-commit` hook automatically, no need to call `make PKG=MSnbase NEWS` manually.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,7 @@
 # MSnbase 2.5
 
 ## Changes in version 2.5.1
-- Update dependencies (see issue [#271](lgatto/MSnbase/issues/271))
+- Update dependencies (see issue [#271](https://github.com/lgatto/MSnbase/issues/271))
 - Replace HCD by ETD in TMT10ETD's name/description
 
 ## Changes in version 2.5.0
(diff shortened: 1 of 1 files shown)


sgibb:
Unfortunately `git pull` doesn't update this hook. Do you know a way to distribute hooks? Otherwise we maybe should create a `hooks` directory and provide a `setup.sh` that copies the hooks into `.git/hooks`. Or do I overlook something

lgatto:
Indeed, the `.git` directory isn't under version control. I have the same issue with [branch-specific `.gitignore` files](https://lgatto.github.io/branch-specific-gitignore/). What about an new `Maker` recipe: `make pre-commit-news`?

sgibb:
That would be complicated. Because we don't want to overwrite the whole file. What about a `hook` directory and a script/`maker` recipe to create symlinks?

lgatto:
> What about a `hook` directory and a script/`maker` recipe to create symlinks?

So calling `make pre-commit-news` would check if there already is a `pre-commit` file, and either creates one or adds a new hook? Yes, that sounds good to me!

sgibb:
To be honest that was not was I meant :wink:. I thought about a `hook` directory for `MSnbase` and a `maker` recipe that creates symlinks. But IMHO it would be better to create a `hook` directory for `maker` with common hooks (e.g. a hook for convert the links in `NEWS.md`, for creating the `NEWS` automatically, a `README.Rmd` hook (that looks whether the `.Rmd` is newer than the corresponding `.md`) and `pkgdown` hook and so on). Ideally you would just run `make PKG=MSnbase create-hook pre-commit-NEWS.md` and it will set it up. I will create an issue against `maker`.

PR #284, opened 2017-12-08
msnset se coercions

arnesmits:
First step towards an MSnSet to SE coercions.


R/msnset2se.R
```diff
@@ -0,0 +1,48 @@
...
+#'
+#' \code{se2msnset} generates an MSnSet object from an SummarizedExperiment object.
+#'
+#' @param se SummarizedExperiment,
+#' Object which will be turned into an MSnSet object.
+#' @return MSnSet object.
+#' @export
+se2msnset <- function(se) {
+  # Check input
+  assertthat::assert_that(inherits(se, "SummarizedExperiment"))
```
sgibb: Currently we don't use `assertthat` for evaluation of input variables. While it is very convenient it don't think we are accepting a new dependency for just an easy `if (inherits(se, "SummarizedExperiment"))` check. BTW: `assertthat` is not listed in the `DESCRIPTION` file.

R/msnset2se.R
```diff
@@ -0,0 +1,48 @@
...
+
+  # Extract expression, feature and pheno data
+  raw <- SummarizedExperiment::assay(se)
```
sgibb: Currently we also have no dependency on `SummarizedExperiment`. That means you have to add `SummarizedExperiment` to the `DESCRIPTION` file. I would suggest to add it into `Suggests:` and use `if (requireNamespace("SummarizedExperiment"))` here to check whether it is available or not.

R/msnset2se.R
```diff
@@ -0,0 +1,48 @@
...
+  feat_data <- data.frame(SummarizedExperiment::rowData(se))
+  rownames(feat_data) <- se@NAMES
```
sgibb: You should not access slots directly. Why not use `names(se)` here? I am not very familiar with `SummarizedExperiment` yet but are you sure that `rownames(rowData(se))` is never set?

R/msnset2se.R
```diff
@@ -0,0 +1,48 @@
...
+  pheno_data <- data.frame(SummarizedExperiment::colData(se))
+
+  # Generate MSnSet
+  msnset <- MSnbase::MSnSet(exprs = as.matrix(raw),
```
sgibb: There is no need to use `MSnbase::` because there are no other `MSnSet`s around and the namespace of the package is always used first.

R/msnset2se.R
```diff
@@ -0,0 +1,48 @@
...
+                            pData = Biobase::AnnotatedDataFrame(pheno_data),
+                            fData = Biobase::AnnotatedDataFrame(feat_data))
```
sgibb: We don't need `Biobase::` as well. We are depending on `Biobase` so `AnnotatedDataFrame` is available without `::`.

R/msnset2se.R
```diff
@@ -0,0 +1,48 @@
...
+  return(msnset)
```
sgibb: `return` isn't necessary here. The last object created in the function would be returned automatically. So `msnset <- ` is also not necessary.

R/msnset2se.R
```diff
@@ -0,0 +1,48 @@
...
+#'
+#' \code{msnset2se} generates an SummarizedExperiment object from an MSnSet object.
+#'
+#' @param msnset MSnSet object,
+#' Object which will be turned into an SummarizedExperiment object.
+#' @return SummarizedExperiment object.
+#' @export
+msnset2se <- function(msnset) {
+  # Check input
+  assertthat::assert_that(inherits(msnset, "MSnSet"))
```
sgibb: See my comment above regarding `assertthat`.

R/msnset2se.R
```diff
@@ -0,0 +1,48 @@
...
+
+  # Extract assay, rowData and colData
+  raw <- Biobase::exprs(msnset)
+  row_data <- Biobase::fData(msnset)
+  col_data <- Biobase::pData(msnset)
```
sgibb: We don't need `Biobase::` here as well.

R/msnset2se.R
```diff
@@ -0,0 +1,48 @@
...
+
+  # Generate SE
+  se <- SummarizedExperiment::SummarizedExperiment(assays = as.matrix(raw),
+                                                   rowData = row_data,
+                                                   colData = col_data)
+  return(se)
```
sgibb: `return` not needed.

R/msnset2se.R
```diff
@@ -0,0 +1,48 @@
...
+  # Check input
+  assertthat::assert_that(inherits(msnset, "MSnSet"))
+
+  # Extract assay, rowData and colData
+  raw <- Biobase::exprs(msnset)
+  row_data <- Biobase::fData(msnset)
+  col_data <- Biobase::pData(msnset)
+
+  # Generate SE
+  se <- SummarizedExperiment::SummarizedExperiment(assays = as.matrix(raw),
```
sgibb: If `SummarizedExperiment` was attached by `requireNamespace` (which is missing here) the `::` operator isn't needed.

R/msnset2se.R
```diff
@@ -0,0 +1,48 @@
...
+#' Object which will be turned into an MSnSet object.
+#' @return MSnSet object.
+#' @export
+se2msnset <- function(se) {
+  # Check input
+  assertthat::assert_that(inherits(se, "SummarizedExperiment"))
+
+  # Extract expression, feature and pheno data
+  raw <- SummarizedExperiment::assay(se)
+  feat_data <- data.frame(SummarizedExperiment::rowData(se))
```
sgibb: Please don't use *snake_case*. We try to follow the [bioc coding style](https://bioconductor.org/developers/how-to/coding-style/) and use *camelCase*

R/msnset2se.R
arnesmits: I added requireNamespace, however I still need the :: operator.

R/msnset2se.R
```diff
@@ -5,22 +5,23 @@
...
 #' Object which will be turned into an MSnSet object.
 #' @return MSnSet object.
-#' @export
 se2msnset <- function(se) {
   # Check input
-  assertthat::assert_that(inherits(se, "SummarizedExperiment"))
+  stopifnot(inherits(se, "SummarizedExperiment"))
+	
+	if(!requireNamespace("SummarizedExperiment")) {
+		stop("The SummarizedExperiment package is required for SummarizedExperiment to MSnSet conversion.")
```
sgibb: Sorry for being petty but could you please use just 4 spaces for indentation (you used ~8) and keep lines shorter than 80 characters?

R/msnset2se.R
sgibb: The `::` should not necessary. You only need it if you have two functions with the same name in two imported packages. Beside adding `SummarizedExperiment` to the `DESCRIPTION` file you have to import the class and the methods you use via `NAMESPACE`: `importClassesFrom(SummarizedExperiment, SummarizedExperiment)` and `importFrom(SummarizedExperiment, assay, colData, rowData)`.

R/msnset2se.R
```diff
@@ -5,22 +5,23 @@
...
+  featData <- data.frame(SummarizedExperiment::rowData(se), row.names = names(se))
+  phenoData <- data.frame(SummarizedExperiment::colData(se))
 
   # Generate MSnSet
-  msnset <- MSnbase::MSnSet(exprs = as.matrix(raw),
-                            pData = Biobase::AnnotatedDataFrame(pheno_data),
-                            fData = Biobase::AnnotatedDataFrame(feat_data))
-  return(msnset)
+  MSnSet(exprs = as.matrix(raw),
+  			 pData = AnnotatedDataFrame(phenoData),
```
sgibb: Please fix the indentation here as well.

R/msnset2se.R
```diff
@@ -30,19 +31,33 @@ se2msnset <- function(se) {
...
-  return(se)
+  SummarizedExperiment::SummarizedExperiment(assays = as.matrix(raw),
+  																					 rowData = rowData,
+  																					 colData = colData)
 }
+
+setAs("MSnSet", 
+			"SummarizedExperiment",
+			function(from) {
+				msnset2se(from)
```
sgibb: The functions are very short and not used anywhere else. That's why I don't see the point of an additional internal function. Please put the function body into the `setAs` method.

tests/testthat/test_msnset_se_coercions.R
```diff
@@ -21,6 +21,6 @@ test_that("se2msnset conversion", {
 	expect_equal(colnames(msnset_back), colnames(se))
 	expect_equal(rownames(msnset_back), rownames(se))
 	expect_equal(Biobase::exprs(msnset_back), SummarizedExperiment::assay(se))
-	expect_equal(Biobase::fData(msnset_back), data.frame(SummarizedExperiment::rowData(se), row.names = se@NAMES))
+	expect_equal(Biobase::fData(msnset_back), data.frame(SummarizedExperiment::rowData(se), row.names = names(se)))
```
sgibb: I missed this one before. Why do you compare against `msnset_back` and not directly against `msnset`? And wouldn't be `expect_equal(msnset, as(se, "MSnSet"))` enough?

And the `::` shouldn't be needed here.

tests/testthat/test_msnset_se_coercions.R
```diff
@@ -8,7 +8,7 @@ test_that("msnset2se conversion", {
 	expect_equal(colnames(msnset), colnames(se))
 	expect_equal(rownames(msnset), rownames(se))
 	expect_equal(Biobase::exprs(msnset), SummarizedExperiment::assay(se))
-	expect_equal(Biobase::fData(msnset), data.frame(SummarizedExperiment::rowData(se), row.names = se@NAMES))
+	expect_equal(Biobase::fData(msnset), data.frame(SummarizedExperiment::rowData(se), row.names = names(se)))
```
sgibb: The `::` shouldn't be needed here.

R/msnset2se.R
lgatto: @arnesmits - I believe you use use `se2msnset` and `msnset2se` in your package and possibly mention them in documentation or vignette. I think we can keep them, at least in a first instance, as they might already be used here and there. But the official `as` syntax should probably he preferred.

lgatto:
Also, @arnesmits, please add yourself as a contributor in the DESCRIPTION file.

R/msnset2se.R
sgibb: I am against exporting `se2msnset` and `msnset2se`:

1. We prefer the `as` syntax.
2. `x2y` is uncommon.
2. If we add it we would teach the user a "wrong" approach.
3. If a second package (`MSnbase`) supports them they become more common and seem more "correct".
4. To deprecate/defunct a function is effort (we have to remember it for more than one release cycle, documentation, ...) and we maybe won't ever do it.

R/msnset2se.R
sgibb: BTW: if you use `setAs` instead of an internal function the `stopifnot(inherits(se, "SummarizedExperiment"))` is superfluous.

R/msnset2se.R
lgatto: @arnesmits - are you ok with marking the old forms as Defunct and suggest to use the standard `as` form, so we can redirect users that might already have used them to a more standard form?

arnesmits: I agree with both of you that the common _as_ syntax would be preferable. The old form is mentioned in the documentation of my package, although it is not required in the default workflow. I will mark it as Defunct and suggest the _as_ form, once this is incorporated in MSnbase.

tests/testthat/test_msnset_se_coercions.R
arnesmits: One issue with the current coercion is that it only converts the assayData, phenoData and featureData. The other slots, such as qual, processingData, experimentData, protocolData and annotation are _MSnSet_ specific. The _SummarizedExperiment_ supports a simple list as metadata, so one solution could be to parse all _MSnSet_ specific slots into a list which is stored in the _SummarizedExperiment_ metadata and vice versa for _SE_ to _MSnSet_ conversion. The question is whether it is necessary to transfer this additional information and whether it is preferred to store this information in a simple list?

lgatto: I will push the necessary `as(x, "list")` to `MSnbase` in the coming days, so that you can add these slots to the `SummarizedExperiment`'s `metadata` slot. Will update this comment with more instructions when the code becomes available.

lgatto: I have added the three `as` methods in the latest `MSnbase` on github. There is also `MSnbase:::.reduce_list` that will remove empty elements or those with and empty character (`""`) - `MSnbase:::.reduce_list(as(processingData(x), "list"))`, so as to avoid adding missing data. The names of the elements could possibly be prefixed with `MSnbase` to make it explicit where these metadata originate from.

lgatto:
@sgibb - I moved that package to `Imports` to fix the error, but I'm happy to leave it `Suggests` to avoid the hard dependency.

sgibb:
Yes, but I am afraid that is not possible. I don't know if there is a way to provide a method for a class without importing this class.

sgibb:
BTW: it is not a warning but a message and I already get the message:
`in method for ‘coerce’ with signature ‘"IBSpectra","MSnSet"’: no definition for class “IBSpectra”`
So we could move `SummarizedExperiment` to `Suggests:` and add a few more messages about missing class definitions to the `MSnbase` startup ... 😞

lgatto:
Yes, I can live with the message.

arnesmits:
I agree that avoiding a hard dependency is preferred, so changed this back.

tests/testthat/test_msnset_se_coercions.R
arnesmits: @lgatto: Thanks. In the end, I didn't use these functions as I could store the _data.frame_, _MSnProcess_, _MIAPE_ and _AnnotatedDataFrame_ objects in the metadata _list_ of SummarizedExperiment.

The current implementation is a bit 'MSnSet-focused', because it will correctly coerce all MSnSet slots into SE and back to MSnSet. However, SE metadata that does not fit into one of the MSnSet slots will be lost in the SE to MSnSet coercion.

lgatto: The issue with this is that once you have an object created with 
```r
se <- as(msnset, "SummarizedExperiment")
```
you will still need `MSnbase` to use `se` because of the `MSnProcess` and `MIAPE` objects in the metadata (`AnnotatedDataFrame` is defined in `Biobase`, which is a dependency of `SummarizedExperiment`). Having that assumption simplifies the code, but it limits the portability of that `se` object. @arnesmits - what is your typical use case?

sgibb:
@arnesmits After moving `SummarizedExperiments` from `Imports` to `Suggests` you also have to remove the entries from `NAMESPACE`. Could you please add some information to the `MSnSet-class` manual page.

tests/testthat/test_msnset_se_coercions.R
lgatto: I will work on this next week, and only transfer some information as a list to the `SummarizedExperiment` metadata slot, and document the behaviour in the manual page.

lgatto:
Done now. See [`./R/coerce.R`](https://github.com/lgatto/MSnbase/blob/master/R/coerce.R#L75-L155) and [`./man/MSnSet-class.Rd`](https://github.com/lgatto/MSnbase/blob/master/man/MSnSet-class.Rd#L248-L252) and [here](https://github.com/lgatto/MSnbase/blob/master/man/MSnSet-class.Rd#L441-L449) for details. The unit tests are in [`test_coerce.R](https://github.com/lgatto/MSnbase/blob/master/tests/testthat/test_coerce.R#L24-L61).

arnesmits:
@lgatto @sgibb: Thanks a lot for all your work!

Issue #283, opened 2017-12-05
Required fvarLabels


lgatto:
Allow to only create a subset of required feature variables from the header. this should be possible on object construction (as an argument to `readMSData` and as a function).

jotsetung:
Extending on it:
+ `readMSData` allows to manually specify wanted `fvarLabels` - by default use all columns returned by `mzR::header`.
+ `reduceFvarLabels` (or `reduceFeatureData`) to remove eventually un-necessary columns from the `featureData`.

jotsetung:
Note: this will only work for `OnDiskMSnExp`.

jotsetung:
I would define a variable with the absolutely necessary column names for the `featureData`:
```r
.OnDiskMSnExpReqFvarLabels <- c("fileIdx", "spIdx", "acquisitionNum",
                                "retentionTime", "msLevel")
```
Is there anything missing from your side @lgatto @sgibb ?

lgatto:
> Note: this will only work for `OnDiskMSnExp`.

Yes, that makes sense, but could we name the vector `.MSnExpReqFvarLabels`.

Regarding the function name to do this, what about `filterFvarData`/`filterFeatureData` (it does filtering along the feature variable columns, just like other filter functions do along the lines) or `keepFvarData`/`keepFeatureData`, or `selectFeatureData`. I am not convinced about *reduce*, that already has a meaning in the package. 

There happen to already exist a [`selectFeatureData` function](https://github.com/lgatto/MSnbase/blob/master/R/functions-MSnSet.R#L237), which allows the user to select which ones to keep if no `fcol` is missing (default). This function could probably be moved out of the `functions-MSnSet` file it currently lives in (`utils.R` is probably a good place), and amended so that if `object` is an `MSnExp` and `fcol` is missing, `fcol` is set to `MSnExpReqFvarLabels`.

sgibb:
Considering #282 I would also like to keep `precursorScanNum` per default (especially for `OnDiskMSnExp` objects).

> Note: this will only work for OnDiskMSnExp.

But the function `filterFeatureData` should work for `MSnExp` as well. Because at least `FilterString` etc. will be part of the `featureData` of an `MSnExp` object, or?

jotsetung:
OK, added `precursorScanNum` to the required `fvarLabels`.

[...]

PR #285, opened 2017-12-11
Add support for OnDiskMSnExp in selectFeatureData

jotsetung:
- Add support for OnDiskMSnExp in selectFeatureData function (issue #283).
- Add requiredFvarLabels function to return the minimal required columns in the
  featureData of an object.
- Fix Spectrum1 and Spectrum2 constructor functions to accept also missing
  values for some columns/attributes.
- Add related unit tests and documentation.


R/functions-Spectrum1.R
```diff
@@ -102,60 +103,63 @@ Spectra1_mz_sorted <- function(peaksCount = NULL, rt = numeric(),
...
+    empty_log <- as.logical(empty_int)
     ## Now match all of the lengths to the length of nvalues.
-    if (length(peaksCount) == 0)
+    if (!length(peaksCount))
         peaksCount <- nvalues
     ## rt
-    if (length(rt) == 0){
-        rt <- rep(NA_integer_, nvals)
+    if (!length(rt)) {
+        rt <- empty_int
```
sgibb: Are all this `if(!length(...))` statements really necessary? I have no clue about the input values. Would it be possible to set the default argument of this constructor to e.g. `rt=rep_len(NA_integer_, length(nvalues))`?

R/utils.R
```diff
@@ -1248,3 +1248,111 @@ setMethod("reduce", "data.frame",
...
+##' ## Reduce the feature data data.frame to the required columns only
+##' od <- selectFeatureData(od, fcol = requiredFvarLabels(class(od)))
+##' fvarLabels(od)
+selectFeatureData <- function(object,
+                              graphics = TRUE,
+                              fcol) {
+    if (missing(fcol)) {
+        if (graphics) {
+            if (!requireNamespace("shiny"))
+                warning("The shiny package is required to use the graphical interface.")
```
sgibb: IMHO this should be `if (graphics && !requireNamespace("shiny"))`. Currently a warning is thrown if `shiny` is missing. Nevertheless the `.selectShinyFeatureData` is executed and will fail.

@lgatto `MSnbase` really suggests `shiny` just for `fcol` selection? `MSnbase` has already a lot of dependencies. Starting a browser and build a webpage for selecting some columns of a `DataFrame` is a little bit weird. I would suggest to remove this `shiny` part.

R/utils.R
```diff
@@ -1248,3 +1248,111 @@ setMethod("reduce", "data.frame",
...
+#' @return For `requiredFvarLabels`: `character` with the required feature
+#'     variable names.
+#' 
+#' @md
+requiredFvarLabels <- function(x = c("OnDiskMSnExp", "MSnExp", "MSnSet")) {
+    x <- match.arg(x)
+    varl <- character()
+    if (x == "OnDiskMSnExp")
+        varl <- .MSnExpReqFvarLabels
+    varl
```
sgibb: `varl` needed? What about
```r
if (match.arg(x) == "OnDiskMSnExp")
    .MSnExpReqFvarLabels
else
    character()
```

jotsetung:
never ignoring any of your suggestions @sgibb ! thanks for reviewing.

R/functions-OnDiskMSnExp.R
```diff
@@ -88,15 +88,20 @@ subsetFeatureDataBy <- function(fd, index = NULL, scanIdx = NULL,
...
     ## originalPeaksCount.
-    reqCols <- c("fileIdx", "spIdx", "acquisitionNum",
-                 "retentionTime", "polarity", "msLevel",
-                 "totIonCurrent", "originalPeaksCount",
-                 "centroided")
+    ## reqCols <- c("fileIdx", "spIdx", "acquisitionNum",
+    ##              "retentionTime", "polarity", "msLevel",
+    ##              "totIonCurrent", "originalPeaksCount",
+    ##              "centroided")
+    reqCols <- .MSnExpReqFvarLabels
```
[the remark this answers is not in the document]
jotsetung: nope, don't need `reqCols` - will remove.

R/functions-Spectrum1.R
```diff
@@ -102,60 +103,63 @@ Spectra1_mz_sorted <- function(peaksCount = NULL, rt = numeric(),
     if (length(mz) != length(intensity))
         stop("Lengths of 'mz' and 'intensity' do not match!")
     nvals <- length(nvalues)
+    empty_int <- rep(NA_integer_, nvals)
+    empty_num <- as.numeric(empty_int)
+    empty_log <- as.logical(empty_int)
```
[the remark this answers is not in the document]
jotsetung: 😄 no way - `snake_case` rules. I'll actually replace em all with 🐫 since the rest of the function uses also only `CamelCase`.

R/functions-Spectrum1.R
jotsetung: I know, it's ugly this way. The *problem* is that the upstream function uses e.g. `fdata$polarity` to pass the polarity column from the feature data to the function. If the `featureData` has been reduced and column `"polarity"` was removed, `fdata$polarity` is `NULL` and `NULL` is then passed with the `polarity` parameter.

R/utils.R
jotsetung: We could move `shiny` from `Suggests` to `Enhances` and use your suggested `requireNamespace`. Most users will have `shiny` anyway installed.

R/utils.R
jotsetung: 👍

R/utils.R
lgatto: In some applications (mainly with `MSnSet` for spatial proteomics), large feature tables are really annoying, and a shiny app to easily select them is handy (we even have such a tab in our `pRolocGUI` apps). 

@jotsetung - I don't think that `Enhances` is correct - it would imply that `MSnbase` enhances `shiny` (the meaning of `Enhances` has been confused and is confusing). `Suggests` is the right place.

jotsetung: OK, I'll put it back to `Suggests`.

lgatto:
@jotsetung - I think you forgot to `git add fdata-selection.R`.

jotsetung:
Aaaah, sorry for that! I did push it now - too late

PR #288, opened 2017-12-16
Add filterPrecursorScan,MSnExp-method.

### LINKED ISSUE #282: Link MS2 to MS1, MS3 to MS2 and MSn to MS(n-1) spectra
Last week @pavel-shliaha ask me to check whether a new neutral loss method for phosphopeptides is working as expected. Therefore it was important to find out which MS3 spectrum was triggered by which MS2. I did that by matching the precursorMz used for the MS2 scan from the *filterString* (the MS3 scan filterString contains the precursor for MS2 and MS3; that was one of the reasons for implementing it into `mzR`):
```r
fData(ms2)$filterString[31:36]
# [1] "ITMS + c NSI t d Full ms2 893.7355@cid35.00 [241.0000-1798.0000]"
# [2] "ITMS + c NSI t d Full ms2 1078.4039@cid35.00 [291.0000-2000.0000]"
# [3] "ITMS + c NSI t d Full ms2 993.4641@cid35.00 [268.0000-1997.0000]"
# [4] "ITMS + c NSI t d Full ms2 584.0798@cid35.00 [155.0000-1179.0000]"
# [5] "ITMS + c NSI t d Full ms2 933.4402@cid35.00 [251.0000-1877.0000]"
# [6] "ITMS + c NSI t d Full ms2 978.3182@cid35.00 [264.0000-1967.0000]"
fData(ms3)$filterString[1:5]
# [1] "FTMS + p NSI d Full ms3 933.4402@cid35.00 915.4110@hcd30.00 [100.0000-2000.0000]"
# [2] "FTMS + p NSI d Full ms3 850.0895@cid35.00 752.1000@hcd30.00 [100.0000-2000.0000]"
# [3] "FTMS + p NSI d Full ms3 644.5992@cid35.00 603.6380@hcd30.00 [100.0000-2000.0000]"
# [4] "FTMS + p NSI d Full ms3 699.0558@cid35.00 666.1801@hcd30.00 [100.0000-2000.0000]"
# [5] "FTMS + p NSI d Full ms3 636.8463@cid35.00 619.0060@hcd30.00 [100.0000-2000.0000]"
```
I did the matching as follows:
```r
## create data.frames for MS2 and MS3 with scanId and precursorMz
ms2df <- data.frame(scanId=scanIndex(ms2),
                    pcCid=gsub(".*ms2 *([0-9.]+@cid[0-9.]+) .*", "\\1",
                               fData(ms2)$filterString),
                    stringsAsFactors=FALSE)
ms3df <- data.frame(scanId=scanIndex(ms3),
                    pcCid=gsub(".*ms3 *([0-9.]+@cid[0-9.]+) .*", "\\1",
                               fData(ms3)$filterString),
                    stringsAsFactors=FALSE)

## merge them by precursorMz
nldf <- merge(ms3df, ms2df,
              all.x=TRUE, all.y=FALSE,
              sort=FALSE, by="pcCid", suffixes=c(".ms3", ".ms2"))

## calculate differences in scan Ids
nldf$delta <- nldf$scanId.ms3 - nldf$scanId.ms2

## remove all scans where the MS2 was acquired after the MS3
nldf <- nldf[nldf$delta > 0,]

## keep just the first (earliest) MS3 (if there are duplicated ones)
nldf <- nldf[order(nldf$scanId.ms3, nldf$delta),]
nldf <- nldf[!duplicated(nldf$scanId.ms3),]
```

See the whole example at https://rawgit.com/sgibb/talk-odense-20171127/master/01-neutral-loss.html#compare-nl-trigger (unfortunately it doesn't contain the dataset because it is 2.2 Gb but @pavel-shliaha promise me that he would share this dataset if needed).

While this procedure isn't very complicated we should provide a function or something else to allow an easier reference for each MSn to its corresponding MS(n-1). 
I am not sure that the *filterString* approach will always work. Maybe we have to go through all precursorMz or some other information. Also I am not sure whether it would be enough to add another column to the `featureData` slot (e.g. `parentMsScanId`) or we need to add it to the `Spectrum` class.

Do you have any better ideas? Do you thing we need such a kind of reference?

sgibb:
This PR closes #282. Briefly it adds a method to filter linked spectra (MS1 and its corresponding MS2 and if available their corresponding MS3 and so on). The information is available in the `featureData` via the `precursorScanNum` column. 

I use two `for` loops for the up/downward search in the hierarchy because recursion is slow in R and constructing the family tree would be more time consuming than the filtering.

I am wondering how to handle the argument `scanIndex` in `filterPrecursorScan`. In fact it is not the `scanIndex` but the `acquisitionNum`. Should I change the name or is there a way to translate between the two different indices?


R/methods-filters.R
```diff
@@ -164,3 +164,16 @@ setMethod("filterEmptySpectra", "MSnExp",
               object <- logging(object, msg)
               object
           })
+
+setMethod("filterPrecursorScan", "MSnExp",
+          function(object, scanIndex, ...) {
+            object <- object[.filterSpectraHierarchy(fData(object), scanIndex)]
```
jotsetung: This method will never work for `MSnExp` objects, because they don't have the required columns in the feature data. Any plans to implement it for `MSnExp` objects? A simple helper function that iterates through the `Spectrum` objects to extract the `acquisitionNum` and `precursorScanNum` and returns a `data.frame` might do it.

lgatto:
This all looks good. I would like to do two things before merging

1. get my hands on an file with MS3 data for some more testing (`msdata/proteomics/MS3TMT10_01022016_32917-33481.mzML.gz` suffers from the Xcalibur bug documented in #283)
2. add a section in the vignette that shows how to use the code and visualise to respective peaks.

@sgibb - Re 1. above, do you know if it's possible to re-process a raw file to fix the missing precursor data? Otherwise, I will try to get a new file (and hope we have a recent Xcalibur version).

R/methods-filters.R
sgibb: Currently I don't have any plans to add such a function (because most `MSnExp` won't have more than one `msLevel`) but I used `MSnExp` to provide this option.

sgibb:
@lgatto Sorry, but I never used `Xcalibur` myself. If you can't find the correct `Xcalibur` version @pavel-shliaha could convert the files for us.

Putting this code into the vignette would be great (but you are right, we need an example file).

What about the argument name `scanIndex`?

lgatto:
@sgibb - I can get a file processed with Xcallibur 3.x tomorrow. 

> What about the argument name `scanIndex`?

Indeed, this is confusing (I had to double check myself when testing yesterday). I think it should be `acquisitionNum`, to make it match with the feature variable.

lgatto:
@jotsetung - is this something that needs to be pushed to Bioc quickly?

R/methods-filters.R
jotsetung: Right, forgot about that.

jotsetung:
Re `scanIndex`: if I recall correctly, `scanIndex` is the index of the spectrum entry within the file, which does not have to be identical to the `acquisitionNum` which represents the order of the spectra as detected by the machine. `pwiz` determines the `acquisitionNum` and the `precursorScanNum` both from the ID of the spectrum in its respective entry in the mzML. This is obviously vendor specific, so in many instances the `scanIndex` and `acquisitionNum` are identical.

lgatto:
@sgibb - Thank you for this, great feature. I am close this but keeping issue #282 open until I add some test data (in `msdata`) and examples.

lgatto:
I have now added an example that uses an MS3 TMT11.

Issue #286, opened 2017-12-12
Implement a `readChromData` function


jotsetung:
Implement a `readChromData` (or `readChromatograms`) function to read chromatogram data from an mzML file.
This relates to https://github.com/sneumann/mzR/issues/141 in `mzR`.

jotsetung:
@wilsontom and @lgatto, a question: if you run an SRM/MRM experiment and have multiple files (e.g. for multiple samples), you will always have chromatograms for the same ions in each file, right?

If that's the case, we can use the `Chromatograms` object to represent a SRM or MRM experiment. The `Chromatograms` organizes `Chromatogram` objects in a 2-dimensional array, columns being samples, rows chromatograms for the same m/z value (or m/z value range).

If there is the possibility that two files have chromatograms for completely different (and unrelated) ions (m/z) we have to implement a different object that would be more like the `MSnExp` that organizes `Spectrum` objects in one dimension.

PR #293, opened 2018-01-10
Read chrom data

jotsetung:
PR adds the `readSRMData` function (issue #286) that allows to read mzML files with chromatographic data from SRM/MRM experiments along with documentation, unit tests and vignette updates.


R/readChromData.R
```diff
@@ -0,0 +1,246 @@
...
+
+#' Convert the mzML polarity information into a character (+, -, NA)
+#' representing the polarity
+#'
+#' @noRd
+#'
+#' @md
+#'
+#' @author Johannes Rainer
+.polarity_char <- function(x) {
```
sgibb: Is there a unit test for this function? Is the output `c(NA, "-", "+")` expected for `c(-1, 0, 1)`? (see comment above)

R/methods-Chromatograms.R
```diff
@@ -430,3 +430,15 @@ setMethod("mz", "Chromatograms", function(object) {
     .mz_chromatograms(object, mz = "mz")
 })
 
+#' @rdname Chromatograms-class
+#'
+#' @description
+#'
+#' \code{polarity}: returns the polarity of the scans/chromatograms: `1`,
+#' `0` or `-1` for positive, negative or unknown polarity.
```
sgibb: Is polarity really encoded as 1 for positiv, 0 for negative and -1 for unknown?
```
[Term]
id: MS:1000465
name: scan polarity
def: "Relative orientation of the electromagnetic field during the selection and detection of ions in the mass spectrometer." [PSI:MS]
relationship: part_of MS:1000441 ! scan
```
I have never seen anything else than a "positive scan" (and never touched any SRM/MRM data) but IHMO it would be logic to encode a "negative scan" as minus one (and unknown as zero or something different).

jotsetung: I know - having +1 for positive, -1 for negative and NA for unknown would make more sense. I believe it comes from the `ramp` c++ library - and `mzR` uses that for all backends (`Ramp`, `pwiz`) - there has been an issue on that because I was wondering about that encoding too (can't find the issue - could be in `mzR` or `xcms`). 
The `header` function in `mzR` definitely returns this encoding and that's also the value we add to the `Spectrum` objects in `readMSData`.

R/readChromData.R
jotsetung: No, no unit test (yet - I'll add) and yes, the function is expected to return `"+"` for `1`, `"-"` for `0` and `NA` for `-1`. This function is only used to create IDs for SRM/MRM chromatograms similar to the one that is reported in the `mzML` file.

R/readChromData.R
```diff
@@ -0,0 +1,246 @@
...
+    fd <- do.call(rbind, x)
+    if (missing(cols))
+        cols <- colnames(fd)
+    else {
+        if (!all(cols %in% colnames(fd)))
+            stop("All columns specified with 'cols' have to be present in the",
+                 " data.frames")
+    }
+    ## Make first unique elements
+    fd <- unique(fd[, cols, drop = FALSE])
```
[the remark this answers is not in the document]
jotsetung: Thanks for the suggestion - I'll check it (also in terms of performance).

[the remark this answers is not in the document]
jotsetung: Ah, you didn't like the snake_case variable names in the original code, that's why you want to change it ;)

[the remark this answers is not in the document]
jotsetung: yes, your's is also hard to understand - I failed to make it running on the example data set - could you please push a `.combine_data.frame2` function so I could test?

R/readChromData.R
jotsetung: I've added now the unit test for the function (https://github.com/lgatto/MSnbase/pull/293/commits/824766ee268a70fb5c3b2a848ae76be8207916e0)

tests/testthat/test_readChromData.R
```diff
@@ -0,0 +1,62 @@
...
+    expect_equal(mrm$sample, 1:2)
+})
+
+test_that(".combine_data.frame works", {
+    A <- data.frame(a = c(1, 1, 2, 3, 4, 5, 6), b = c(2, 2, 3, 4, 5, 5, 6))
+    B <- data.frame(a = c(1, 3, 3, 3, 3, 4, 5), b = c(2, 4, 4, 4, 5, 5, 5))
+    C <- data.frame(a = c(3, 3, 4, 4), b = c(4, 5, 5, 5))
+    
+    .combine_data.frame(list(A, B, C))
+    exp <- data.frame(a = c(1, 1, 2, 3, 4, 5, 6))
```
sgibb: Here you missed an `expect_equal(.combine_data.frame(list(A, B, C)), exp)` call.

R/readChromData.R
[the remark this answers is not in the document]
sgibb: > could you please push a .combine_data.frame2 function so I could test?

Done.

> yes, your's is also hard to understand

In fact I just create the `data.frame` id to avoid one loop. Could be slightly faster. It works at least for your `list(A, B, C)` and `list(D, E)` test cases. BTW in the unit test you create `list(A, B, C)` and `exp` but never compare the results.

> Ah, you didn't like the snake_case variable names in the original code, that's why you want to change it ;)

You got me :wink:

tests/testthat/test_readChromData.R
jotsetung: thanks for pointing this out. Have added it now.

lgatto:
Pushed to Bioc

Issue #299, opened 2018-01-22
missing phenoData<- for OnDiskMSnExp and AnnotatedDataFrame


jotsetung:
The `phenoData<-` method for `OnDiskMSnExp` and `AnnotatedDataFrame` or `NAnnotatedDataFrame` is missing.

jotsetung:
Closing since this seems to have been fixed

PR #301, opened 2018-01-22
Add phenoData<- method (issue #299)

jotsetung:
- Add phenoData<-,pSet method.
- Add documentation and related unit tests.


R/methods-pSet.R
```diff
@@ -546,10 +546,29 @@ setReplaceMethod("$", "pSet", function(x, name, value) {
 })
 
 setReplaceMethod("pData", "pSet", function(object, value) {
+    if (!is.data.frame(value))
```
lgatto: Is this necessary? Trying `pData(obj) <- x` with anything else than a `data.frame` leads to an error
```
Error in (function (classes, fdef, mtable)  : 
  unable to find an inherited method for function ‘pData<-’ for signature ‘"NAnnotatedDataFrame", "not.a.data.frame"’
```

R/methods-pSet.R
```diff
@@ -546,10 +546,29 @@ setReplaceMethod("$", "pSet", function(x, name, value) {
...
+        stop("'value' has to be a 'data.frame'")
     pData(object@phenoData) <- value
     object
 })
 
+setReplaceMethod("phenoData", "pSet", function(object, value) {
+    ## Check input and convert to NAnnotatedDataFrame if neccessary
+    if (is.data.frame(value)) {
```
lgatto: I think it would be better that `phenoData<-` worked with an `AnnotatedDataFrame` and a `NAnnotatedDataFrame`, and that `pData<-` was reserved for a `data.frame`. I think this is closer to the semantic of the accessorts.

R/methods-pSet.R
jotsetung: I know - I just wanted to throw a more user-friendly error. There might be users not understanding what the *default* error message means.

R/methods-pSet.R
jotsetung: OK

R/methods-pSet.R
lgatto: Ok then.

Issue #308, opened 2018-02-07
Bug in setAs(x, "MSnExp")


jotsetung:
`setAs(x, "MSnExp")` seems to not transfer all of the required data to the resulting `MSnExp` object:
```r
> library(MSnbase)
> f <- c(system.file("microtofq/MM14.mzML", package = "msdata"),
       system.file("microtofq/MM8.mzML", package = "msdata")) 
> od <- readMSData(f, mode = "onDisk")
> im <- as(od, "MSnExp")
> validObject(od)
[1] TRUE
> validObject(im)
[1] TRUE
> validObject(filterFile(od, 1))
[1] TRUE
> validObject(filterFile(od, 2))
[1] TRUE
> validObject(filterFile(im, 1))
[1] TRUE
> validObject(filterFile(im, 2))
Error in validObject(x) : 
  invalid class "MSnExp" object: Mismatch of files in assayData and processingData.
```

jotsetung:
There is a conceptional problem:
+ `setAs` copies all slots from the `OnDiskMSnExp` over to the `MSnExp`, also the `fData`. 
+ Feature data of `OnDiskMSnExp` contains a column named `"fileIdx"`, feature data of `MSnExp` does not.
+ `filterFile,OnDiskMSnExp` ensures that the the entries in the `"fileIdx"` are updated to represent the correct indices. `filterFile,MSnExp` is agnostic of this column and does thus also not update it.
+ Wham: `validObject,pSet` fails.

2 Solutions:
1) amend the `validObject` method.
2) ensure that column `"fileIdx"` is not copied in the coercion.

I'd go for option 2) because 1) might cause other problems.

lgatto:
> I'd go for option 2) because 1) might cause other problems.

Yes, I agree. Looking into the PR now.

PR #309, opened 2018-02-07
Fix issue #308

jotsetung:
Fixes issue #308 by removing column `"fileIdx"` from the `fData` of the `MSnExp` object returned from `as(x, "MSnExp")`.


lgatto:
Pushed to Bioc.

PR #312, opened 2018-02-16
Add extracMsData and plotMsData functions

jotsetung:
- Add extractMsData and plotMsData function to extract MS data as a data.frame
  and visualize the data.
- Fix removePeaks,Spectrum after recent changes in IRanges (introduced in
  Bioconductor 3.7).

The `extractMsData` and `plotMsData` are two functions I had defined in `xcms` but they might be better placed in `MSnbase` because they provide some more low level functionality. I will need these then later in the examples and vignette related to the *centroiding* that I am currently developing in the `centroiding` branch.


R/AllGenerics.R
```diff
@@ -147,3 +147,6 @@ setGeneric("splitByFile", function(object, f, ...) standardGeneric("splitByFile"
 setGeneric("productMz", function(object, ...) standardGeneric("productMz"))
 ## setGeneric("aggregationFun", function(object, ...)
 ##     standardGeneric("aggregationFun"))
+
+setGeneric("extractMsData", function(object, ...)
```
jotsetung: I'm using here a method instead of a function, because in `xcms` I have a `extractMsData,XCMSnExp` method that has some additional functionality (ensures that adjusted retention times are used if present).

lgatto:
I haven't looked into it in details nor run the code, but my first question is, why `extractMsData` and not `as(, "data.frame")`. The `setAs` functions don't support extra arguments (in your case `rt`, `mz` and `msLevel`, but couldn't this be done using appropriate filtering functions `filterRt`, `filterMz` and `filterMsLevel`? 

Also, re using `as`, see `as(Spectrum, "data.frame")` to possibly use within the `MSnExp` coercion method.

lgatto:
I haven't looked at the output of `plotMsData`, but why have it use a `data.frame` and not an `MSnExp`?

jotsetung:
Re using `as(Spectrum, "data.frame")`: this one does not return the `rt` values that I need. Also, I'm directly accessing the slots in the function and am not using the `mz` and `intensity` methods. I've found that calling the methods takes, if applied to a large number of spectra, considerably longer because R has to look up the method.

I do however like the idea of a `as(<MSnExp>, "data.frame")`. This should then return a `data.frame` with columns `"file"`, `"rt"`, `"mz"` and `"i"`. This could replace the proposed `extractMsData` method.

jotsetung:
Re `plotMsData`: the reason was that I used it along the `extractMsData`, so, extract data with `extractMsData`, which returns a `data.frame` and plot that data with `plotMsData`. In the end calling it on a `MSnExp` would/should also work - ideally by specifying a `file` argument to plot the data for a single file. The name of the function is also not the best I have to admit.

jotsetung:
Thinking it over I think I'll reduce the pull request to a single `setAs` method that converts a `MSnExp` to a `data.frame` and drop the `plotMsData` altogether.

lgatto:
A `plot,MSnExp` would be a useful addition, if you have something general and useful enough. If you think the `plotMsData` function could play such a role, could you paste the produced figure in here.

jotsetung:
That is what `plotMsData` produces:

kc_3_alanine_3

so, it's the m/z against retention time at bottom (intensity color-coded) and on top of that the base peak chromatogram (highest signal per spectrum). Makes mostly sense for MS1 and if you have LC-MS data.

PR #317, opened 2018-03-12
Updated grammar in "Description" field

chasemc:


lgatto:
Thank you! I bet initially had the present participles and then changed it to nouns with the prepositions :-/

Issue #321, opened 2018-03-21
Disable CV parameter "MS:-1" for writeMSData


jotsetung:
This is related to https://github.com/sneumann/mzR/issues/151 . Apparently, while `proteowizard` allows to use `MS:-1` for an unknown CV this seems not to be allowed in an mzML file.

Amend thus the `.guesSoftwareProcessing` to not return a CV param if it is unknown.

jotsetung:
That was a little too fast - I need the `"MS:-1"` for know to differentiate between MS CV IDs for the software and the processings (we can change this to use the *real* MS CV term for `MSnbase` once that is included).

Will have to revert this -

lgatto:
Ok, go ahead.

jotsetung:
I made a pull request addressing this. 

We should keep the issue open until the `MSnbase` has its own MS CV term ID and the replace `"MS:-1"` with that.

lgatto:
I think this can be closed now, right?

PR #324, opened 2018-03-23
Ensure that a MS CV term ID for the software is added

jotsetung:
Ensure that a MS CV term ID for the software is added. This is currently `"MS:-1"`, since `MSnbase` does not have its own term, but should in future be replaced with the one from `MSnbase` (See issue #321 for more info).


jotsetung:
I'm merging this, since it is not affecting any other functionality than `writeMSData`.

PR #328, opened 2018-03-29
Group m/z values from consecutive scans based on sqrt(mz) values

jotsetung:
This PR adds the parameter `timeDomain` to `combineSpectra`, `combineSpectraMovingWindow` and `estimateMzScattering` and performs the definition of m/z - intensity pairs from consecutive scans that should be merged (i.e. that are from the same ion) based on the `sqrt(mz)` instead of `mz` values. The rationale is that QTOF MS profile data is most likely collected in regular time intervals (m/z steps, or m/z resolution) based on a timing circuit. m/z values are calculated then using the relationship `t = k * sqrt(mz)`. As a result, the spacing between m/z values will not be constant over the full m/z range. Performing calculations on `sqrt(mz)` minimizes this effect.

So far I did not experience any difference between `timeDomain = TRUE` and `timeDomain = FALSE`, but `timeDomain = TRUE` should be more robust against the above effect in some situations.


R/functions-MSnExp.R
```diff
@@ -373,10 +373,20 @@ removeReporters_MSnExp <- function(object, reporters = NULL,
...
+#' scattering of m/z values, but the scattering of `sqrt(mz)` values.
+#' 
 #' @param x `MSnExp` or `OnDiskMSnExp` object.
 #'
 #' @param halfWindowSize `integer(1)` defining the half window size for the
 #'     moving window to combine consecutive spectra.
+#'
+#' @param timeDomain `logical(1)` whether m/z scattering should be estimated
+#'     on `mz` (`timeDomain = FALSE`) or `sqrt(mz)` (`timeDomain = TRUE`)
+#'     values. See [combineSpectraMovingWindow()] for deteils on this
```
sgibb: typo: "deteils" -> "details".

R/functions-MSnExp.R
```diff
@@ -449,6 +459,17 @@ estimateMzScattering <- function(x, halfWindowSize = 1L) {
 #' 
 #' See [combineSpectra()] for details.
 #'
+#' Parameter `timeDomain`: by default, m/z-intensity pairs from consecutive
+#' scans to be aggregated are defined based on the square root of the m/z
+#' values. This is because it is highly likely that in all QTOF MS instruments
+#' data is collected based on a timing circuit (with a certain variance) and
+#' m/z values are later derived based on the relationship `t = k * sqrt(m/z)`.
+#' differences between individual m/z values will thus be dependent on the
```
sgibb: typo: "differences" -> "Differences".

R/functions-Spectrum.R
```diff
@@ -851,7 +851,15 @@ descendPeak <- function(mz, intensity, peakIdx = NULL, signalPercentage = 33,
 #' All m/z values with a difference smaller than this value are combined to
 #' a m/z group.
 #' Intensities and m/z values falling within each of these m/z groups are
-#' aggregated using the `intensity_fun` and `mz_fun`, respectively.
+#' aggregated using the `intensity_fun` and `mz_fun`, respectively. It is
+#' highly likely that all QTOF profile data is collected with a timing circuit
+#' that collects data poings with regular intervals of time that are then later
```
sgibb: typo: "poings" -> "points".

SiggiSmara:
Maybe not the right place and I apologize beforehand if I am hijacking the pull request. I am more than happy to move this to an issue on your suggestions. This could also be just my newb-ness with R and I am doing something silly.

I was wanting to try this feature out but ran into an error loading a file with readMSData.  To reproduce, I installed R devel from source and then installed bioconductor packages making sure useDevel was on. Once mzR, MSnbase and xcms all installed via biocLite with no problems I installed MSnbase from github via devtools using mz_scatter as ref.
Then I ran this:
```R
library(msdata)
library(xcms)

mzdatapath <- file.path(find.package("msdata"), "sciex")
files <- list.files(mzdatapath, recursive = TRUE, full.names = TRUE)

data_profile <- readMSData(files[1], mode = "onDisk")
```
And I get the following error:

> Error in value[[3L]](cond) : duplicate 'row.names' are not allowed
>   AnnotatedDataFrame 'initialize' could not update varMetadata:
>   perhaps pData and varMetadata are inconsistent?
> In addition: Warning message:
> non-unique value when setting 'row.names': ‘centroided’ 
> 

I also then tried downgrading MSnbase to the version in bioconductor devel as well as installing the current master from github. Both with the same result.

jotsetung:
@SiggiSmara the error you get is because you most likely have the newest `mzR` that reads also centroiding information from mzML files. I have another pull request in `MSnbase` (https://github.com/lgatto/MSnbase/pull/329) that fixes this problem.

Workaround for now is to manually downgrade `mzR` to version 1.13.5

jotsetung:
Thanks @sgibb for the review 👍

R/functions-MSnExp.R
```diff
@@ -519,22 +540,25 @@ combineSpectraMovingWindow <- function(x, halfWindowSize = 1L,
                                        mzFun = base::mean,
                                        intensityFun = base::mean,
                                        mzd = NULL,
-                                       BPPARAM = bpparam()){
+                                       BPPARAM = bpparam(),
```
lgatto: Wouldn't it be better to keep `BPPARAM` as the last parameter? This is its usual position in all functions.

R/functions-Spectrum.R
```diff
@@ -882,7 +898,7 @@ descendPeak <- function(mz, intensity, peakIdx = NULL, signalPercentage = 33,
 #' of m/z and intensity pairs than the spectrum with index `main` in `x`, also
 #' all other related information is taken from this spectrum.
 #'
-#' @author Johannes Rainer
+#' @author Johannes Rainer, Sigurdur Smarason
```
lgatto: I Sigurdur Smarason @SiggiSmara has the co-author of the new code, could you also add hime as a contributor (`role = "ctb"`) to the `DESCRIPTION` file, please.

R/functions-MSnExp.R
jotsetung: Yes, agreed. I'll change.

R/functions-Spectrum.R
jotsetung: OK

SiggiSmara: :smiley: thanks! @lgatto

PR #349, opened 2018-06-26
Fix bug in `robustSummary()`

adder:
When samplenames of input are not alphabetically ordered, the order of the `robustSummary()` output is incorrect.

Model.matrix orders the ellements in a categorical variable alphabetically.
So the coefficients of rlm are also ordered alphabetically.
This order should be again reversed in the output into the same order as the samplesnames of input.

The proposed code change fixes this.

### DIFF
--- a/R/combineFeatures.R
+++ b/R/combineFeatures.R
@@ -321,10 +321,10 @@ robustSummary <- function(e, residuals = FALSE, ...) {
     ## This will be needed for NUSE-type of quality control, but will
     ## need to check for missing data as below.
     ## se <- unique(summary(fit)$coefficients[sampleid, 'Std. Error'])
-
-    ## Put NA for the samples without any expression value
-    present <- as.logical(colSums(p))
-    res <- rep(NA, length(present))
-    res[present] <- fit$coefficients[sampleid]
-    res
+  
+    ## Take the sample coefficients ( = summarised expression values)
+    coef = fit$coefficients[sampleid]
+    ## Sort the sample coefficients in the same way as the samplenames of expression matrix
+    ## Puts NA for the samples without any expression value  
+    coef[paste0('sample',colnames(e))]
 }


jotsetung:
Just a note: you could avoid this alphabetical re-ordering from the beginning by defining the factors providing also the levels:
```r
## Instead of
sample <- rep(colnames(e), each = nrow(e))[p]

## Use
sample <- factor(rep(colnames(e), each = nrow(e))[p], levels = colnames(e))
```
there should be no need to sort them afterwords.
you could do the same for `feature`.

adder:
Yes, you are offcourse right. That would be a more elegant way to handle the sample ordering.
One remark: I think you should also remove the samples that have missing intensities from `levels`
Else `model.matrix` makes to many collumns.
```r
## Instead of
sample <- factor(rep(colnames(e), each = nrow(e))[p], levels = colnames(e))

## Use
sample <- factor(rep(colnames(e), each = nrow(e))[p], levels = colnames(e)[p])
```

As a side, note that my suggested patch also handels missing values in the output (samples with no intensity = NA) while reducing the involved lines of code from 4 to 2 :)

R/combineFeatures.R
```diff
@@ -321,10 +321,10 @@ robustSummary <- function(e, residuals = FALSE, ...) {
...
-    res <- rep(NA, length(present))
-    res[present] <- fit$coefficients[sampleid]
-    res
+  
+    ## Take the sample coefficients ( = summarised expression values)
+    coef = fit$coefficients[sampleid]
+    ## Sort the sample coefficients in the same way as the samplenames of expression matrix
+    ## Puts NA for the samples without any expression value  
+    coef[paste0('sample',colnames(e))]
 }
```
sgibb: Why you don't replace line 318 `fit <- MASS::rlm(X, expression, ...)` with
```r
MASS::rlm(X, expression)$coefficients[paste0("sample", colnames(e))]
```
(maybe covered by `unname`) and remove all the following lines?

adder:
@sgibb  Yes this should indeed provide the wanted results in the least amount of lines of codes. 
I'm in favor of this.

@lgatto Not sure how to proceed from here.
This code is already in the latest stable bioconductor release,but this is in my opinion a serious bug.
I only noticed it now because I only used the MSnBase version of robust summary in combination with maxquant and maxquant apparently orders samples alphabetically by default in its output.
But in data from other sources this is of course often not the case.

Can this still fixed somehow in stable bioconductor?
Or only is this only possible in the next release?

greets

lgatto:
> This code is already in the latest stable bioconductor release,but this is in my opinion a serious bug. Can this still fixed somehow in stable bioconductor?

Yes, that is possible and recommended. I will fix the bug in the devel and release. I hope to be able to review the PR tonight.

lgatto:
Sorry for the delay - looking into this now. 

@adder - in general, in such cases I would suggest to also provide a unit test that identifies the bug and verifies that is is fixed.

PR #355, opened 2018-08-06
Fix filterMz for spectra with no non-NA intensities in m/z range

jotsetung:
This fixes issue https://github.com/sneumann/xcms/issues/301, seems the files contained `NA` intensities.


sgibb:
I didn't read the xcms issue. But you said there seem to be `NA` values for **intensities** but you changed the filtering for **mz** values. Is that intended? 
BTW: I like the `which`/`length` approach and often use it myself but is there a reason why you changed the current code and not just added `na.rm` to `sum` here (`if (sum(sel, na.rm=TRUE) == 0)`/`if (!sum(sel, na.rm=TRUE))`).

jotsetung:
You're right, it is the m/z not intensity. I didn't check the actual values and why `NA` m/z are returned.

Ah, you got me. Yes, indeed, the simpler solution would have been to just use `if (sum(sel, na.rm = TRUE) == 0)`. I just didn't think of that, no special reason. You want me to change to that?

jotsetung:
@lgatto , can you please also merge this (or manually change the respective lines) into the release version of MSnbase/Bioconductor?

lgatto:
I have now backported this to release in version 2.6.3.

PR #362, opened 2018-10-10
Internal MGF export functions gain parameter addFields

jotsetung:
Add parameter addFields to internal writeMgfDataFile and writeMgfContent to support addition of arbitrary additional fields in the MGF output file.

### DIFF
--- a/R/readWriteMgfData.R
+++ b/R/readWriteMgfData.R
@@ -20,12 +20,27 @@ setMethod("writeMgfData",
                              verbose = verbose)
           })
 
+#' @param addFields `data.frame` or `matrix` with optional additional
+#'     fields to be added to each spectrum (each column one field).
+#'
+#' @noRd
 writeMgfDataFile <- function(splist, con, COM = NULL, TITLE = NULL,
-                             verbose = isMSnbaseVerbose()) {
+                             verbose = isMSnbaseVerbose(), addFields = NULL) {
   if (class(con) == "character" && file.exists(con)) {
     message("Overwriting ", con, "!")
     unlink(con)
   }
+
+  if (length(addFields)) {
+      if (is.null(dim(addFields)))
+          stop("'addFields' has to be a matrix or data.frame.")
+      else addFields <- as.matrix(addFields, ncol = ncol(addFields))
+      if (is.null(colnames(addFields)))
+          stop("Column names required on 'addFields'.")
+      if (nrow(addFields) != length(splist))
+          stop("nrow of 'addFields' has to match length of 'splist'")
+      haveAddFields <- TRUE
+  } else haveAddFields <- FALSE
 
   con <- file(description = con, open = "at")
   on.exit(close(con))
@@ -45,13 +60,17 @@ writeMgfDataFile <- function(splist, con, COM = NULL, TITLE = NULL,
     if (verbose)
       setTxtProgressBar(pb, i)
 
-    writeMgfContent(splist[[i]], TITLE = NULL, con = con)
+    if (haveAddFields)
+        writeMgfContent(splist[[i]], TITLE = NULL, con = con,
+                        addFields = addFields[i, ])
+    else
+        writeMgfContent(splist[[i]], TITLE = NULL, con = con)
   }
   if (verbose)
     close(pb)
 }
 
-writeMgfContent <- function(sp, TITLE = NULL, con) {
+writeMgfContent <- function(sp, TITLE = NULL, con, addFields = TRUE) {
     .cat <- function(..., file = con, sep = "", append = TRUE) {
         cat(..., file = file, sep = sep, append = append)
     }
@@ -83,6 +102,11 @@ writeMgfContent <- function(sp, TITLE = NULL, con) {
         }
     } else .cat("\nRTINSECONDS=", rtime(sp))
 
+    if (length(addFields)) {
+        if (!is.null(names(addFields)))
+            .cat(paste0("\n", paste(toupper(names(addFields)),
+                                    addFields, sep = "="), collapse = ""))
+    }
 
     .cat("\n", paste(mz(sp), intensity(sp), collapse = "\n"))
     .cat("\nEND IONS\n")
--- a/R/readWriteMgfData.R
+++ b/R/readWriteMgfData.R
@@ -34,7 +34,8 @@ writeMgfDataFile <- function(splist, con, COM = NULL, TITLE = NULL,
   if (length(addFields)) {
       if (length(dim(addFields)) != 2)
           stop("'addFields' has to be a matrix or data.frame.")
-      else addFields <- as.matrix(addFields, ncol = ncol(addFields))
+      if (!is.matrix(addFields))
+          addFields <- do.call(cbind, lapply(addFields, as.character))
       if (is.null(colnames(addFields)))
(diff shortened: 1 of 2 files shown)


R/readWriteMgfData.R
```diff
@@ -20,13 +20,28 @@ setMethod("writeMgfData",
...
 
+  if (length(addFields)) {
+      if (is.null(dim(addFields)))
+          stop("'addFields' has to be a matrix or data.frame.")
+      else addFields <- as.matrix(addFields, ncol = ncol(addFields))
+      if (is.null(colnames(addFields)))
+          stop("Column names required on 'addFields'.")
+      if (nrow(addFields) != length(splist))
+          stop("nrow of 'addFields' has to match length of 'splist'")
+      haveAddFields <- TRUE
```
sgibb: Do you need this additional variable `haveAddFields`? Would not `is.null(addFields)` be enough in line 63 (`if (haveAddFields)`). But even this would not be needed at all (see below).

R/readWriteMgfData.R
```diff
@@ -45,13 +60,17 @@ writeMgfDataFile <- function(splist, con, COM = NULL, TITLE = NULL,
     if (verbose)
       setTxtProgressBar(pb, i)
 
-    writeMgfContent(splist[[i]], TITLE = NULL, con = con)
+    if (haveAddFields)
```
sgibb: Do we really need this `if` statement here. If we reach this point `addFields` is a `matrix` or `NULL`. And you can subset `NULL`:

```r
f <- function(n, addFields=NULL) {
    for (i in seq_len(n)) {
        print(addFields[i, ])
    }
}

f(3)
# NULL
# NULL
# NULL

f(3, matrix(1:6, ncol=2, dimnames=list(NULL, c("FOO", "BAR"))))
# FOO BAR 
#   1   4 
# FOO BAR 
#   2   5 
# FOO BAR 
#   3   6 
```

While it looks a little bit odd/dirty the workhorse function `writeMgfContent` will test for `NULL` itself.

R/readWriteMgfData.R
```diff
@@ -45,13 +60,17 @@ writeMgfDataFile <- function(splist, con, COM = NULL, TITLE = NULL,
...
+        writeMgfContent(splist[[i]], TITLE = NULL, con = con,
+                        addFields = addFields[i, ])
```
sgibb: I know it is not your fault because it is already part of the original code but `TITLE = NULL` overwrites the `TITLE` argument. Seems like a bug nobody found/mentioned yet.

R/readWriteMgfData.R
```diff
@@ -45,13 +60,17 @@ writeMgfDataFile <- function(splist, con, COM = NULL, TITLE = NULL,
...
+    else
+        writeMgfContent(splist[[i]], TITLE = NULL, con = con)
   }
   if (verbose)
     close(pb)
 }
 
-writeMgfContent <- function(sp, TITLE = NULL, con) {
+writeMgfContent <- function(sp, TITLE = NULL, con, addFields = TRUE) {
```
sgibb: I don't understand the default value `TRUE` for `addFields` here.

R/readWriteMgfData.R
```diff
@@ -83,6 +102,11 @@ writeMgfContent <- function(sp, TITLE = NULL, con) {
         }
     } else .cat("\nRTINSECONDS=", rtime(sp))
 
+    if (length(addFields)) {
```
sgibb: Instead of the nested `if` statement: `length(addFields) && !is.null(names(addFields))` or `is.vector(addFields) && !is.null(names(addFields))`?

R/readWriteMgfData.R
```diff
@@ -83,6 +102,11 @@ writeMgfContent <- function(sp, TITLE = NULL, con) {
...
+        if (!is.null(names(addFields)))
+            .cat(paste0("\n", paste(toupper(names(addFields)),
+                                    addFields, sep = "="), collapse = ""))
```
sgibb: `.cat("\n", paste(toupper(names(addFields)), addFields, sep = "=", collapse = "\n"))`

R/readWriteMgfData.R
```diff
@@ -20,13 +20,28 @@ setMethod("writeMgfData",
...
 writeMgfDataFile <- function(splist, con, COM = NULL, TITLE = NULL,
-                             verbose = isMSnbaseVerbose()) {
+                             verbose = isMSnbaseVerbose(), addFields = NULL) {
   if (class(con) == "character" && file.exists(con)) {
     message("Overwriting ", con, "!")
     unlink(con)
   }
 
+  if (length(addFields)) {
+      if (is.null(dim(addFields)))
```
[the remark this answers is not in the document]
jotsetung: Good point @sgibb ! Thanks! I'll fix.

R/readWriteMgfData.R
jotsetung: No, we don't. Usually I add a variable instead of calling `is.null` 2 or more times on the same object in a function. Doesn't apply here, but if the function would be called in a loop this can have a performance effect.
But good point with the `NULL` subsetting! Actually never tried subsetting `NULL` 😄

R/readWriteMgfData.R
jotsetung: 👍

R/readWriteMgfData.R
jotsetung: good point! I'll fix.

R/readWriteMgfData.R
jotsetung: Ups, well, me neither. I'll change to `addFields = NULL`.

R/readWriteMgfData.R
jotsetung: good point. Will change.

R/readWriteMgfData.R
jotsetung: yes, better that way. Thanks.

R/readWriteMgfData.R
```diff
@@ -20,13 +20,28 @@ setMethod("writeMgfData",
...
   if (class(con) == "character" && file.exists(con)) {
     message("Overwriting ", con, "!")
     unlink(con)
   }
 
+  if (length(addFields)) {
+      if (length(dim(addFields)) != 2)
+          stop("'addFields' has to be a matrix or data.frame.")
+      if (!is.matrix(addFields))
+          addFields <- do.call(cbind, lapply(addFields, as.character))
```
sgibb: I really hate `as.matrix` for calling `format` and padding all `character`s with whitespaces. Instead of `do.call` and `lapply` we could use `addFields <- trimws(as.matrix(d))` here.

sgibb: BTW: Internal `as.matrix.data.frame` uses a `for` loop. So depends on your style whether you like the `do.call, cbind, lapply` or `trimws(as.matrix())` approach.

jotsetung: I'd rather take the `do.call, cbind, lapply` avenue - it's faster then the `trimws(as.matrix)`. But thanks for the suggestion.

R/readWriteMgfData.R
jotsetung: Strange. In fact I **did** fix it. Don't know why it does not show above.

lgatto:
:+1:

Issue #201, opened 2017-04-05
speed up package load time


jorainer:
The loading time for `MSnbase` is quite long. I was wondering if we could reduce the package loading time (about 3.3 secs on my computer). This would be specifically helpful for parallel processing using `SnowParam` (only parallel processing available in Windows).

I guess the loading time is a combination of time spent on the imports and on the exports - so I guess there is not much that can be done.

lgatto:
It would be great if we could to that. Maybe worth exploring:
- move some packages from depends to imports in the description
- reduce import to importFrom in the namespace

jotsetung:
I'll play a little around with that and what speed improvement we could get.

sgibb:
We could also review all Depends and Suggests. E.g. `reshape2` is just used for `melt` (which could easily replaced), see #202 for example. (I am not sure whether @lgatto want to merge this now, or want to wait for the next release cycle).

lgatto:
As long as changes are relatively minor, like #202, I am happy to merge now.

jotsetung:
I would rather like to change stuff later - don't want to break things again ;)

PR #368, opened 2018-10-19
Split dependency building and pkg checking on travis

sgibb:
Currently our CI fails on travis because the hard runtime limit for jobs
is 50 minutes and installing all dependencies of MSnbase takes a lot of
more time.

image

This commit introduces multiple changes:
- Use 2 cores for package building/installation.
- Split building of dependencies and package checking into different
  jobs which allows us to use 2x50 minutes. The build job builds all dependencies and create a cache which is used by the check job.
- Automatically deploy of `pkgdown` based documentation to github pages the master branch.

@lgatto if you want to use the automatic deployment you have to create a [**P**ersonal **A**ccess **T**oken](https://help.github.com/articles/creating-a-personal-access-token-for-the-command-line/) with the `public_repo`  scope/permission and register it in the [travis' settings](https://docs.travis-ci.com/user/environment-variables#Defining-Variables-in-Repository-Settings) with the name `GITHUB_PAT`

We could create even more jobs to reduce the check time a little bit (currently after cache building it takes ~ 35 min to run travis but using three check jobs ("examples", "vignettes", "test") we could go down to ~ 25 min:

**2 jobs (build + check, current PR)**

travis1

**5 jobs (build + 3 check + codecov)** (the codecov job failed because I accidentally used `rscript` instead of `Rscript`)

tavis2

@lgatto If you wish I could easily extend this PR to the version with 5 jobs.

BTW: 
- I add `skip_on_travis()` to `tests/testthat/test_MSmap.R` on my forked repository because downloading files from `AnnotationHub` always failed (maybe we should consider it to add it here, too).
- `MSnbase` needs more than 250 packages to build. Maybe we could remove a few, e.g. `shiny` is just used for one function (could `selectFeatureData` move to `pRolocGUI`?), same for `XML` which is just used by `.isCentroidedFromFile` and `mzID` both could be replaced by `mzR`/`xml2` (the latter is downloaded anyway by other dependencies), `affy`/`affyio` just for `MAplot`, `gplots` just for `heatmap.2` in the vignette, ... it is somehow connected to #201 but would just improve the build time. Should we open an issue and try to remove some build dependencies or just keep status quo?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -69,9 +69,7 @@ jobs:
 # In general the r-travis script automatically dump the logs after failure but
 # because of a bug in travis the output is often truncated. See
 # https://github.com/travis-ci/travis-ci/issues/6018
-# (Adding `sleep 10` in after_script as proposed by some people didn't work for
-# me.)
-after_failure: dump_error_logs
+after_failure: sleep 10
 
 notifications:
   email:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,47 +1,57 @@
+# full script at https://github.com/travis-ci/travis-build/blob/master/lib/travis/build/script/r.rb
 language: r
 r: bioc-devel
+bioc_check: true
 cache: packages
-sudo: false       # use container based build system
+sudo: required
 warnings_are_errors: true
 dist: trusty
 
-# Set CXX1X for R-devel, as R-devel does not detect CXX1X support for gcc 4.6.3,
-# this was causing mzR installation to fail
-# see https://bugs.r-project.org/bugzilla3/show_bug.cgi?id=17189
-# workaround stolen from https://github.com/hadley/devtools/blob/1ce84b04568ff7846c3da754f28e7e22a23c8737/.travis.yml#L23-L26
-before_install:
-  - if [[ "$TRAVIS_R_VERSION_STRING" = 'bioc-devel' ]]; then mkdir ~/.R && echo 'CXX1X=g++ -std=c++0x -g -O2 -fPIC' > ~/.R/Makevars; fi
-
+# unfortunately apt packages are not cached
+# we could use these build dependencies just in the "Build" stage and
+# install their runtime libraries counter parts in "Check" but that would make
+# the script more complicated and saves just a few seconds
 addons:
   apt:
     packages:
       - libnetcdf-dev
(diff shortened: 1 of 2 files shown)


lgatto:
Great, thanks @sgibb. I have created a PAT and added it to the travis setting.
I wouldn't mind 5 jobs, if you are happy to do it and think it's worth it.

I would prefer to keep all tests, as much as possible.  

I will need to check if `.isCentroidedFromFile` is still needed, but migrating it to `xml2` would be good anyway. But not sure about `mzID` - I think it is important to verify some id data validity against `mzR` But I indeed think that we could re-implement (nicer) MAplots to get rid of `affy`. We still need a dependency to plot the heatmap in the vignette though.

But yes, let's open a issue to keep track of the dependencies and try to reduce them.

sgibb:
I added `skip_on_travis` to `tests/testthat/test_MSmap.R`. Somehow `AnnotationHub` could not download the file (even when the VM on travis has internet access). Sometimes it is working, e.g.: https://travis-ci.org/sgibb/MSnbase/builds/444428390
But I rerun this last job multiple times without a single success (we just need a single successful download because the `~/.AnnotationHub` is cached, too).

You can simply revert it with `git revert 7b7c32d6293e13d281c85e18b7867df0386f46fc`.

IMHO this could be merged to master now.

Please note: I recognized that `pkgdown::build_site` fails because of:
```
<simpleError in add_classes_to_exports(ns = nsenv, package = package, exports = exports,     nsInfo = nsInfo): in ‘MSnbase’ methods for export not found: reduce>
There were 50 or more warnings (use warnings() to see the first 50)
Error in add_classes_to_exports(ns = nsenv, package = package, exports = exports,  : 
  in ‘MSnbase’ methods for export not found: reduce
```

I am not sure how to solve this (without exporting `reduce`, seems to be related to [pkgload](https://github.com/r-lib/pkgload/blob/9b8d9c4286a4bc90fde21f998298a77b8e584461/R/namespace-env.r#L129-L130)

lgatto:
Thanks!

lgatto:
@sgibb - is the pkgdown site build and deployed automatically, in the end?

lgatto:
I think the automatic deploy to `gh-pages` is a nice option. I modified the settings to use the `gh-pages` branch, but http://lgatto.github.io/MSnbase/ throws a 404 error. Not sure if it's only a matter of time, or did I miss something, @sgibb.

sgibb:
I can't see a 404 but the regular pkgdown site. I guess it was just a matter of waiting for the next push from travis.

PR #373, opened 2018-10-26
Add parameter ppm to consensusSpectrum and meanMzInts

jotsetung:
I've added the parameter `ppm` to the `consensusSpectrum` and `meanMzInts` functions that allow to group mass peaks from different spectra into the same *combined* mass peak in a m/z dependent manner (i.e. if their difference in m/z values is smaller than `ppm` of the m/z of the peaks). This will complement the grouping based on a fixed value (parameter `mzd`).

In case you are wondering, I did also remove some non-unicode characters in comments as they were breaking `devtools::document()`.


lgatto:
Also on Bioc.

Issue #370, opened 2018-10-21
Spectra to/from MSnExp


lgatto:
I am wondering if we wouldn't need methods to coerce `MSnExp` objects to/from `Spectra` objects. One example I am going to implement soon, after combining spectra (see #366), is that these new data will need to be quantified and serialised to `mzML`.

@jotsetung @sgibb - what do you think?

jotsetung:
I'd like to keep the `Spectra` agnostic of where it comes from (e.g. `mzML` or a database). Wouldn't be a `combineSpectra,MSnExp` what you're looking for? You combine the spectra (get an `MSnExp` back), call quantify on that and you then save the `MSnExp` with the `writeMSData` back to `mzML`.

Getting a `Spectra` from an `MSnExp` is rather simple: `Spectra(spectra(<MSnExp>))`.

jotsetung:
I know that `Spectra(spectra` is awkward. For me `spectra` returning a `Spectra` is the cleanest solution (possibly in addition with an `as(<MSnExp>, "Spectra")`.

jotsetung:
hm, my previous comment disappeared...

@sgibb, I know, `Spectra(spectra` is awkward, but that's presently the way how you can convert an `MSnExp` to a `Spectra`. I would opt for `spectra` returing a `Spectra`. The question is what to do with the `fData`, pass that as `mcols` to the `Spectra`? For `OnDiskMSnExp` that can result in quite a heavy `Spectra` object. Also, should `spectra` be identical to `as(<MSnExp>, "Spectra")`? My feeling is that it might be better to return with `spectra` a `Spectra` without all feature data to keep that more lightweight, and to use `as(<MSnExp>, "Spectra")` to pass along also all feature data as `mcols`. On the other hand it might be more intuitive to have also `spectra` return a `Spectra` with feature data as metadata columns.

lgatto:
The `as(from, to)` syntax is general across many object types, and hence I think it would be nice to have `as(<MSnExp>, "Spectra")` and `as(<Spectra>, "MSnExp")`, as it does exactly what's on the tin. And similarly, having `as(<Spectra>, "list")` would to get a plain `list`. I am ok with redundancy and also having constructors `Spectra(<MSnExp>)` and `MSnExp(<Spectra>)`.

As for `mcols`, yes, it always gets passed along to/from `fData`.

jotsetung:
I've added the related conversion methods in https://github.com/lgatto/MSnbase/pull/378

Issue #354, opened 2018-07-02
Implement clean,Chromatograms method


jotsetung:
Implement the `clean,Chromatograms` method.

PR #376, opened 2018-11-05
Add .editorconfig

sgibb:
This PR is just a suggestion, so fell free to close without merging.

It adds a `.editorconfig` file to the repository. [EditorConfig](https://editorconfig.org/) is a platform and **editor**-independent way of defining editor settings/code styles for different files, e.g. tab instead of spaces, indentation by 4 spaces, line ending, ...

I often have the problem that my `nvim` removes spaces at the end of lines (and it seems that your emacs doesn't do this) which obscure my commits. In general I use `git add -p` and just select the lines where the content really changed, ignoring spaces (sometimes I forget this but I can't find a commit that is showing this, yet). Also different settings of indentation can result in errors, e.g.: https://github.com/lgatto/pRoloc/commit/0be04a3 (https://travis-ci.org/lgatto/pRoloc/requests)

More information:
- EditorConfig website: https://editorconfig.org
- Supported settings: https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties

Each develop needs to install the `editorconfig` C program (debian: `apt-get install editorconfig`) and the editor-specific plugin:
- emacs: https://github.com/editorconfig/editorconfig-emacs#readme
- (n)vim: https://github.com/editorconfig/editorconfig-vim#readme
- github (if you edit/look files in the browser) has native editorconfig support.

@lgatto, @jotsetung what do you think?

EDIT: `editorconfig` is a program written in C not in python.

PR #378, opened 2018-11-07
Add some more Spectra methods

jotsetung:
Add `filterMsLevel,Spectra`, `coerce,Spectra,list` and `coerce,Spectra,MSnExp` (issue #370) and `clean,Chromatograms` (issue #354).


R/DataClasses.R
```diff
@@ -523,9 +523,9 @@ setClass("ProcessingStep",
 #'     example, a \code{Chromatogram} for a SRM transition 273 -> 153 will have
 #'     a \code{@precursorMz = c(273, 273)} and a
 #'     \code{@productMz = c(153, 153)}.
-#' 
+#'
```
sgibb: JFTR: first result of the [EditorConfig](https://editorconfig.org) plugin. As I mentioned in #376 these additional spaces at the line ending happened to me all the time (and I always use `git add -p` to exclude them from the commit). In a few commits we would have edited all files and won't be annoyed by trailing whitespaces anymore. Alternatively we could run `for i in *.R; do sed -i 's#[[:space:]]*$##g' ${i}; done` to get rid of them in just one commit.

jotsetung: I don't mind if they get fixed successively. I like your editorconfig addon 👍

jotsetung:
No change requests at all @sgibb ? for real?

R/methods-Spectra.R
```diff
@@ -514,6 +523,44 @@ setMethod("combineSpectra", "Spectra", function(object, fcol,
...
+    if (!length(names(from)))
+        names(from) <- as.character(seq_len(length(from)))
+    if (is.null(mcols(from)))
+        fd <- AnnotatedDataFrame(data.frame(spectrum = seq_len(length(from)),
+                                            row.names = names(from)))
+    else
+        fd <- AnnotatedDataFrame(as.data.frame(mcols(from)))
+    fromf <- as.character(unique(fromFile(from)))
+    process <- new("MSnProcess", files = fromf,
+                   processing = paste("Data converted from Spectra:", date()))
```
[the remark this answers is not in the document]
jotsetung: I wasn't even aware of the `"[", date(), "]"` - I used `readMSData` as *template*.

R/methods-MSnExp.R
```diff
@@ -565,3 +565,15 @@ setAs("MSnExp", "data.frame", function(from) {
...
+
+setAs("MSnExp", "Spectra", function(from) {
+    fdta <- fData(from)
+    red_cn <- c("fileIdx", "spIdx", "smoothed", "seqNum", "acquisitionNum",
+                "msLevel", "polarity", "originalPeaksCount", "totIonCurrent",
+                "retentionTime", "basePeakMZ", "basePeakIntensity",
+                "collisionEnergy", "ionisationEnergy", "precursorScanNum",
+                "precursorMZ", "precursorCharge", "precursorIntensity",
+                "mergedScan", "centroided", "spectrum")
+    fdta <- fdta[, !colnames(fdta) %in% red_cn, drop = FALSE]
```
lgatto: Why aren't all feature variables passed along during coercion? I would prefer to keep all of them by default and have a `selectFeatureData` function to select those to keep.

lgatto: Indeed, thank you. Is there already an accessor for these (`centroided,Spectra`, ...)? The implementation will be identical to the corresponding in memory `MSnExp` methods.

jotsetung: Yes, all methods to access slot values for `Spectrum` objects are also implemented for `Spectra` objects (in *methods-Spectra.R*), returning always a vector of length equal to `Spectra` with the corresponding slot values.

Issue #380, opened 2018-11-15
Impute KNN sets seed


const-ae:
Hi Laurent,

first thanks for the great package, it makes dealing with MS data much simpler.

I was just exploring the different methods for imputing missing values, when I discovered that to my surprise `impute(msnset, "knn")` fixes the seed value.

``` r
suppressPackageStartupMessages(library(MSnbase))
data(msnset)
# Other methods show the expected behavior
for(i in 1:10){
  tmp <- MSnbase::impute(msnset, "zero")
  print(rnorm(1))
}
#> [1] 2.077938
#> [1] -0.3538495
#> [1] 1.269365
#> [1] 0.1198152
#> [1] -0.07994367
#> [1] -1.107776
#> [1] 0.5004011
#> [1] -0.7553048
#> [1] -1.225024
#> [1] 0.2958373

# Fixes the seed, so no more randomness afterwards
for(i in 1:10){
  tmp <- MSnbase::impute(msnset, "knn")
  print(rnorm(1))
}
#> [1] -0.668527
#> [1] -0.668527
#> [1] -0.668527
#> [1] -0.668527
#> [1] -0.668527
#> [1] -0.668527
#> [1] -0.668527
#> [1] -0.668527
#> [1] -0.668527
#> [1] -0.668527
```

<sup>Created on 2018-11-15 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

I saw that internally it is actually not your package that calls `set.seed()`, but the [`impute::impute.knn()`](https://www.rdocumentation.org/packages/impute/versions/1.46.0/topics/impute.knn) method, but the method returns the state of the random number generator. So I was wondering if you could reset the rng state after the method call? If you are currently busy I could also try to make a small PR.

Best, Constantin

lgatto:
Thank you for this, could catch. I wouldn't mind a PR.

PR #381, opened 2018-11-15
Fix issue #380

const-ae:
- Reset .Random.seed after impute::impute.knn is called

### DIFF
--- a/R/imputation.R
+++ b/R/imputation.R
@@ -26,7 +26,11 @@ setMethod("impute", "MSnSet",
               ## imputaton methods
               ##
               if (method == "knn") {
-                  .eset <- impute.knn(exprs(object), ...)$data
+                  imp_res <- impute.knn(exprs(object), ...)
+                  .eset <- imp_res$data
+                  if(! is.null(imp_res$rng.state)){
+                    .Random.seed <<- imp_res$rng.state
+                  }
                   exprs(object) <- .eset
               } else if (method == "nbavg") {
                   message("Assuming values are ordered.")
--- a/tests/testthat/test_impute.R
+++ b/tests/testthat/test_impute.R
@@ -86,3 +86,16 @@ test_that("nbavg methods", {
     expect_true(exprs(xx[5, 2]) == 10)
     expect_true(exprs(xx[4, 3]) == 14)
 })
+
+
+test_that("seed is not set by knn imputation method" {
+
+  data(naset, package = "MSnbase")
+  
+  rand <- sapply(1:10, function(idx){
+    xx <- suppressWarnings(impute(naset, "knn"))
+    rnorm(1)
+  })
+  expect_gt(max(rand) - min(rand), 0)
+
+})


const-ae:
Oh yeah, you are right. I fixed the issue in my fork. Do I have to make a new pull request or does Github automatically see the new commit?

R/imputation.R
```diff
@@ -26,7 +26,11 @@ setMethod("impute", "MSnSet",
               ## imputaton methods
               ##
               if (method == "knn") {
-                  .eset <- impute.knn(exprs(object), ...)$data
+                  imp_res <- impute.knn(exprs(object), ...)
+                  .eset <- imp_res$data
+                  if(! is.null(imp_res$rng.state)){
```
sgibb: Just a minor coding style correction (spaces at the wrong place):
```suggestion
                  if(!is.null(imp_res$rng.state)) {
```

R/imputation.R
```diff
@@ -26,7 +26,11 @@ setMethod("impute", "MSnSet",
...
+                    .Random.seed <<- imp_res$rng.state
```
sgibb: Instead of using the `<<-` operator which could have some side-effects because it looks in the parent environments until it finds a variable with the given name (I know it is very unlikely but if a user writes a wrapper function around `impute` and defines a `.Random.seed` variable inside this function it would be a hard to find bug.) I would suggest:
 ```suggestion
                    assign(".Random.seed", imp_res$rng.state, envir=.GlobalEnv)
```

File: R/imputation.R [hunk 3]
```diff
@@ -26,7 +26,11 @@ setMethod("impute", "MSnSet",
...
+                  }
```
sgibb: I am not sure but what about a `else` statement. If `.Random.seed` wasn't set at all. It should not be created here, or?
```suggestion
                  } else {
                      rm(".Random.seed", envir=.GlobalEnv)
                  }
```

R/imputation.R
const-ae: Yes, great point, I forgot that `<<-` might update the wrong variable.

R/imputation.R
const-ae: Yes you are right, I think that is probably the best thing one can do in that case.

R/imputation.R
```diff
@@ -26,7 +26,13 @@ setMethod("impute", "MSnSet",
               ## imputaton methods
               ##
               if (method == "knn") {
-                  .eset <- impute.knn(exprs(object), ...)$data
+                  imp_res <- impute.knn(exprs(object), ...)
+                  .eset <- imp_res$data
+                  if(!is.null(imp_res$rng.state)) {
```
sgibb: Sorry I have overseen it, another space needed:
```suggestion
                  if (!is.null(imp_res$rng.state)) {
```

R/imputation.R
```diff
@@ -26,7 +26,13 @@ setMethod("impute", "MSnSet",
...
+                   assign(".Random.seed", imp_res$rng.state, envir=.GlobalEnv)
```
sgibb: Here are 3 spaces missing (I guess it was my fault, using the web interface instead my editor ...):
```suggestion
                      assign(".Random.seed", imp_res$rng.state, envir=.GlobalEnv)
```

const-ae:
Thanks for helping me make a good PR.
I have updated the NEWS.md file, but do I need to do anything to generate the NEWS file or this done somewhere internally?

const-ae:
Thanks, for the pointer. It looks like a quite useful project, but I have to admit I am still slightly intimidated by all the make stuff ;)

lgatto:
Thank you both for the PR and the code review!

PR #385, opened 2018-11-26
Fix [,Chromatograms to support classes extending Chromatograms

jotsetung:
- Fix [,Chromatograms to support classes extending Chromatograms (issue
  https://github.com/sneumann/xcms/issues/333)

I have a class `XChromatograms` extending `Chromatograms` over in `xcms`, subsetting such classes with `[` returned **always** a `Chromatograms` object. Reason was that the type of class is hardcoded in the `[` method. This pull request fixes this.


jotsetung:
Thanks for merging!

PR #406, opened 2019-01-14
Hdf5 jo

jotsetung:
- Update the `Hdf5MSnExp` object to have one hdf5 file per sample/mzML file
- Implement function to extract data from hdf5 files directly using `.Call` to `rhdf5` C functions.
- Implement `spectrapply,Hdf5MSnExp` that parallelizes data access by file.


R/hdf5.R
```diff
@@ -68,226 +72,133 @@ validHdf5MSnExp <- function(object) {
...
-        for (j in seq_along(fns)) {
-            pb$tick()
-            fn <- fns[j]
-            hdfile <- paste0(file_group, "/", fn)
-            .pks <- pks[[j]]
-            colnames(.pks) <- c("mz", "intensity")
-            rhdf5::h5write(.pks, h5, hdfile, level = comp_level)
-        }
+    fh <- openMSfile(file)
+    hdr <- header(fh)
```
jotsetung: In case you're wondering why we read also the `header` although we don't *need* that data: without the `header` call I got randomly a ProteoWizard error that the lengths of mz and intensity values for a spectrum don't match.

PR #408, opened 2019-01-23
Bring hdf5 branch up-to-date

jotsetung:
- `updateHdf5MSnExp` function to *consolidate*/save any data changes back to the hdf5 files.
- `filterFile,Hdf5MSnExp` to ensure that also the `hdf5file` slot gets updated correctly.
- `convertToHdf5MSnExp` to convert an `OnDiskMSnExp` to a `Hdf5MSnExp` object (instead of `as`, since that does not support required additional arguments.
- unit tests.

### DIFF
--- a/R/hdf5.R
+++ b/R/hdf5.R
@@ -1,42 +1,50 @@
-##' @title The `Hdf5MSnExp` Class for MS Data And Meta-Data
-##' @description The `Hdf5MSnExp` class encapsulates data and
-##'     meta-data for mass spectrometry experiments like the `MSnExp`
-##'     and `OnDiskMSnExp` classes. `Hdf5MSnExp` implements, like
-##'     `OnDiskMSnExp`, an *on disk* model, where the raw data (M/Z
-##'     and intensities) are stored on disk (rather than in memory,
-##'     like `MSnExp` data) using, as the name implies, the hdf5 file
-##'     format. See [`MSnExp`] and [`OnDiskMSnExp`] for a more general
-##'     description of the data, slots and operations.
-##'
-##'     Object of the class are currently created with the
-##'     `readHdf5DiskMSData` function. Later, this backend will be
-##'     included in the main `readMSData` function.
-##'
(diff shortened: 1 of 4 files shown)


R/hdf5.R
```diff
@@ -1,110 +1,207 @@
...
-                        prototype = prototype(
-                            new("VersionedBiobase",
-                                versions = c(classVersion("MSnExp"),
-                                             classVersion("OnDiskMSnExp"),
-                                             Hdf5MSnExp = "0.0.1")),
-                            spectraProcessingQueue = list(),
-                            backend = character()))
-
-validHdf5MSnExp <- function(object) {
+validHdf5MSnExp <- function(object, check_md5 = TRUE) {
```
jotsetung: validity can be checked with and without checking also whether the **content** of the hdf5 files has changed. The latter is performed by default. `validObject()` will perform a *basic* check with `check_md5 = FALSE`.

R/hdf5.R
```diff
@@ -1,110 +1,207 @@
...
+#' x <- convertToHdf5MSnExp(x, hdf5path = paste0(tempdir(), "/2/"))
+#'
+#' MSnbase:::validHdf5MSnExp(x)
+#' x[[12]]
+#'
+#' ## clean up session
+#' file.remove(hdf5FileName(x))
+.Hdf5MSnExp <- setClass("Hdf5MSnExp",
+                        slots = c(hdf5file = "character",
+                                  md5sum = "character"),
```
jotsetung: slot `md5sum` stores the md5 sums of the spectra within a hdf5 file. The same number is stored into the file upon generation of the hdf5 file. Also, `updateHdf5MSnExp` updates both the md5 sum in the file and the one in this slot. If they are not the same, the hdf5 file has been changed.

jotsetung:
I'd rather go for the developments we're doing now in the *backends* and *backends_jo* branches. I will take some stuff from here (hdf5 reading etc) and include it there.

Issue #413, opened 2019-01-25
Synchronize documentation efforts


jorainer:
I'd say we should synchronize the documentation efforts, specifically for the backends. I guess we will simply have a single documentation file for all backends and the roxygen code for that should go to the *Backend-class.R*.

sgibb:
I would suggest an .R file per backend that just adds its specific documentation (using the `@describeIn` roxygen tag), so it results in one large manual page for all backends.

jotsetung:
Agree. Wasn't aware of the `@describeIn` tag - have to look it up.

jotsetung:
I did that with `XChromatogram` and `XChromatograms` classes in `xcms`. I'm using a single rd file for all methods, functions and the classes (always using `@rdname XChromatogram`). To keep it somehow organized I'm using `@section` tags grouping description of functions and methods by topic.

sgibb:
Ok, that's possible but you don't really document the class (would require `\docType{class}` and the slots have to be documented). In fact you misuse the method documentation (`\docType{method}`) and add an alias for `Chromatogram(s)-class` to it. :wink:

jotsetung:
That's right. But that's also something that Martin Morgan did in the `AnnotationFilter` package, so, from that perspective it seems to be acceptable - and users should not use the slots anyway directly.

sgibb:
Ok, if Martin did it, it is more or less official. Let's do it that way.

jotsetung:
Suggestion: we use the same documentation concept for the `MSnExperiment` class (one documentation, method descriptions grouped into sections). 
Also, I'd propose to not export any of the `backend...` methods (and hence documenting them also only internally). The user should not have to deal with these methods.

sgibb:
I agree with both suggestions (same documentation concept for `MSnExperiment`, don't export and document (for the user) the `backend...` methods).

Issue #411, opened 2019-01-25
Implement a new data reading function


jotsetung:
For the new backends we need a new reader. I'd suggest something like `readMSData3 <- function(files, sampleData, backend)`. We should change the name then later...

jotsetung:
So far we're not reading instrument information or similar stuff. Feel free to add that functionality if you think it's important/required @lgatto @sgibb .

Also, I think we should call `backendInitialize` and `backendImportData` on the `backend`.

lgatto:
> So far we're not reading instrument information or similar stuff.

Quick comment on that - all the MIAME/MIAPE business never really caught up, so I suggest we ignore it for now. I will discuss further in another issue.

jotsetung:
Agree - if needed we can add it later. I'd like to keep the `MSnExperiment` as simple and lightweight as possible.

PR #414, opened 2019-01-25
Add readMSnExperiment function

jotsetung:
Add `readMSnExperiment` function (issue #411).

Stuff that I'm not happy: documentation. We should aim for a common way to document the new stuff (see issue #413). For now this is more to add a function to read in header data and to start implementing playing the backends.


R/MSnExperiment-class.R
```diff
@@ -1,6 +1,26 @@
...
+#' @noRd
+.validMSnExperiment <- function(object) {
+    msg <- NULL
+    if (length(object@files)) {
+        if (length(object@files) != nrow(object@sampleData))
+            msg <- c(msg, paste0("Number of files does not match the number",
+                                 " of rows of 'sampleNames'"))
+        if (is.null(object@backend))
+            msg <- c(msg, paste0("If files are present the backend can not be",
+                                 " NULL"))
```
sgibb: I like the way validity checking is done in S4Vectors. TLDR: a function for each slot/aspect that is checked.
https://github.com/Bioconductor/S4Vectors/blob/39bd2363b55a0f18ea65ef9011a823151d708e76/R/DataFrame-class.R#L88-L122

This may makes writing unit tests easier. What do you think?

R/MSnExperiment-class.R
```diff
@@ -57,3 +82,64 @@ setMethod(
...
+    file <- normalizePath(file)
+    if (!missing(sampleData)) {
+        if (is.data.frame(sampleData))
+            sampleData <- DataFrame(sampleData)
+    } else {
+        sampleData <- DataFrame(sampleIdx = seq_along(file))
+    }
+    if (!is.logical(smoothed))
+        stop("smoothed should be a logical")
+    .read_file <- function(z, files, smoothed) {
```
sgibb: I don't like this kind of internal functions (because they are hard to test with unit test).

R/MSnExperiment-class.R
jotsetung: Looks nice - but isn't it too much of an overhead implementing one function for each possible failure? I think it makes mostly sense if one re-uses checks. Let's see how you implement your checks and then I'll eventually adapt 😉

R/MSnExperiment-class.R
jotsetung: If I'm moving the function out of the function I'll also have to find a better name...

R/BackendMzR.R
```diff
@@ -0,0 +1,69 @@
...
+#'
+#' @md
+#'
+#' @noRd
+.spectra_from_file_mzR <- function(file, spectraData) {
+    if (missing(file) || missing(spectraData))
+        stop("Both 'file' and 'spectraData' are required")
+    fl <- .openMSfile(file)
+    on.exit(mzR::close(fl))
+    hdr <- mzR::header(fl, max(spectraData$spIdx))
```
sgibb: Is this needed at all? I am not sure whether I remember correctly but there were some memory leaks if the header wasn't accessed, or?

jotsetung: yep. That's right. Seems to work if we load the header for the spectrum with the highest index.

tests/testthat/test_utils.R
```diff
@@ -598,3 +598,16 @@ test_that("levelIndex works", {
     res <- levelIndex(str, which = "last")
     expect_equal(res, c(a = 4, b = 8, c = 7, d = 11))
 })
+
+test_that(".spectra_from_data works", {
+    spctra <- spectra(tmt_erwinia_on_disk)
```
sgibb: BTW we should be careful writing tests that depend on `MSnExp` and `OnDiskMSnExp`. We need a few to be sure that `MSnExperiment` yield identical results. But if we are going to replace (and deprecate/defunct) `(OnDisk)MSnExp` in the far, far away future we have to rewrite all the tests.

R/BackendMzR.R
jotsetung: Honestly, I don't want to change any of this because it could bring us back to the random reading failuere hell again - I played with this for weeks and that was the minimal acceptable solution.

tests/testthat/test_utils.R
jotsetung: Of course. These are intermediates.

R/BackendMzR.R
sgibb: Ok, I am fine with this and don't want to dig deeper.

PR #418, opened 2019-01-28
Use a general documentation entry for Backends (issue #413)

jotsetung:
Idea is that all documentation for the backends goes to `Backends.Rd` defined in `Backends-class.Rd`. Users have thus all information on one page.


R/BackendMzR-class.R
```diff
@@ -1,41 +1,23 @@
...
-#'
-#' ## TODO: define to create the documentation. Add this to the general
-#' ## backend documentation or have it in its own one.
 setClass("BackendMzR", contains = "Backend")
 
+#' @rdname hidden_aliases
 setMethod("backendReadSpectra", "BackendMzR", function(object, file,
                                                        spectraData, ...,
                                                        BPPARAM=bpparam()) {
+    spd_list <- split.data.frame(spectraData, f = spectraData$fileIdx)
```
sgibb: looking at this it looks really simple and useful. Maybe we should keep all the `BPPARAM` in the backend. Because for `BackendMemory` parallelisation is not useful, for the ondisk backends it may be implemented differently (e.g. SQL needs other parallelisation than mzML/HDF5, ...). So doing parallelisation in the backend could be useful.

jotsetung: 😄 OK, then we're back at the start, but you're right. Also realized that when looking at the `BackendMemory` implementation.

sgibb: Sorry. I thought that the read/write functions would be enough, but as you pointed out in R (as a functional programming language) it would require to load every single spectrum manually (with 1000s of open/close file actions) or read all files at once which takes a lot of memory. So applying the functions inside the backend would be the most memory efficient way.

jotsetung: Agree. So let's go for #422

Issue #420, opened 2019-01-28
Implement spectrapply,MSnExperiment


jotsetung:
Implement the `spectrapply,MSnExperiment` method. This method should parallelize by file (see https://github.com/lgatto/MSnbase/commit/f7673703420ff9b365723495b4509958dd918e50#r32063917) and apply the processing queue on the fly.

PR #421, opened 2019-01-28
Add spectrapply,MSnExperiment (issue #420)

jotsetung:
- Add spectrapply,MSnExperiment method.
- Add .apply_processing_queue function.
- Add related unit tests.


jotsetung:
OK, so you're working on the redifinition? Awesome!

jotsetung:
With the latest commits I moved `@processingQueue` to the `Backend`, removed the `file` parameter from most `backend...` methods and added the `backendSpectrapply` method (for both `BackendMzR` and `BackendMemory` classes.

R/Backend-class.R
```diff
@@ -112,9 +114,16 @@ setClass("Backend",
...
+    if (length(x))
+        if (!all(vapply(x, inherits, logical(1), "ProcessingStep")))
+            return("'processingQueue' should only contain ProcessingStep objects.")
+    NULL
+}
+
 setValidity("Backend", function(object) {
     msg <- .valid.Backend.files(object@files)
-
+    msg <- c(msg, .valid.Backend.processingQueue(object@processingQueue))
```
sgibb: Just use:
```r
msg <- c(
    .valid.Backend.files(object@files),
    .valid.Backend.processingQueue(object@processingQueue)
)
```
instead of 
```r
msg <- c(msg, ...)
msg <- c(msg, ...)
```

R/Backend-class.R
```diff
@@ -166,7 +178,8 @@ setMethod(
     signature="Backend",
     definition=function(object, files, spectraData, ..., BPPARAM=bpparam()) {
     object@files <- normalizePath(files)
-    object
+    if (validObject(object))
```
sgibb: I just wonder if we really need the `if` here (we use it everywhere). `validObject` returns `TRUE` or throws an error. So we could save one if/comparison (I know that wouldn't save much time):
```r
validObject(object)
object
```
What do you think?

R/Backend-class.R
```diff
@@ -266,7 +279,57 @@ setGeneric(
...
+        standardGeneric("backendAddProcessing"),
+    valueClass = "Backend"
+)
+
+setMethod(
+    "backendAddProcessing", "Backend", function(object, spectraData,
+                                                procStep, ...) {
+        if (missing(procStep) || !inherits(procStep, "ProcessingStep"))
+            stop("'procStep' parameter is required and is expected to be of",
+                 " type 'ProcessingStep'")
```
sgibb: The signature could be `c(object="Backend", spectraData="DataFrame", procStep="ProcessingStep")` and we could remove the `if(missing(procStep) ...` (but we lose a clear error message, but because the user will never call any `backend...` methods that doesn't matter).

R/MSnExperiment-class.R
```diff
@@ -146,12 +205,84 @@ readMSnExperiment <- function(file, sampleData, backend = BackendMzR(),
...
+    if (!is.numeric(t) & t != "min")
+        stop("Argument 't' has to be either numeric of 'min'.")
+    if (missing(msLevel.))
+        msLevel. <- base::sort(unique(object@spectraData$msLevel))
+    if (!is.numeric(msLevel.))
+        stop("'msLevel' must be numeric.")
+    object@backend <- backendAddProcessing(
+        object@backend, procStep = ProcessingStep("removePeaks",
+                                                  list(t = t, msLevel. = msLevel.)))
+    validObject(object)
```
sgibb: Here you avoid the `if` call already :+1: 
Coding style suggestion: IMHO it is easier to understand if the last lines are:
```r
object@processing <- c(object@processing, ...)
validObject(object)
object
```
`validObject` should be the last check before returning the new object.

R/MSnExperiment-class.R
```diff
@@ -146,12 +205,84 @@ readMSnExperiment <- function(file, sampleData, backend = BackendMzR(),
...
+    if (!is.logical(all))
+        stop("Argument 'all' must be logical")
+    if (missing(msLevel.))
+        msLevel. <- base::sort(unique(object@spectraData$msLevel))
+    if (!is.numeric(msLevel.))
+        stop("'msLevel' must be numeric.")
+    object@backend <- backendAddProcessing(
+        object@backend, procStep = ProcessingStep("clean",
+                                                  list(all = all, msLevel. = msLevel.)))
+    validObject(object)
```
sgibb: See above ("last check").

R/AllClassUnions.R
```diff
@@ -1,2 +1,3 @@
 setClassUnion("mzIDClasses", c("mzID", "mzIDCollection"))
 setClassUnion("logicalOrNumeric", c("logical", "numeric"))
+setClassUnion("characterOrFunction", c("character", "function"))
```
sgibb: Do we really need this class union? The user never use `ProcessingStep` directly, or? And we could stick with one method (you used `character` before, any use case for `function`?)

tests/testthat/test_Backend.R
```diff
@@ -0,0 +1,5 @@
+test_that(".valid.Backend.processingQueue works", {
```
sgibb: Sorry that I forgot to commit and push my `test_Backend.R` yesterday.

R/AllClassUnions.R
jotsetung: I would like to keep this version. I want to be able to define also a `ProcessingStep` like `ProcessingStep(FUN = function(z, t) z * t, ARGS = list(t = 5))`, i.e. pass along functions, not only function names (provided as `character`).

R/MSnExperiment-class.R
jotsetung: 👍

R/Backend-class.R
jotsetung: makes sense.

R/Backend-class.R
jotsetung: 👍

jotsetung: No need for this in the new code.

R/Backend-class.R
jotsetung: Did that now in the new commit.

R/MSnExperiment-class.R
jotsetung: Did that.

tests/testthat/test_Backend.R
jotsetung: 👍

jotsetung:
See https://github.com/lgatto/MSnbase/issues/422#issuecomment-458919399 for a description of the latest changes.

R/MSnExperiment-class.R
```diff
@@ -142,16 +221,114 @@ readMSnExperiment <- function(file, sampleData, backend = BackendMzR(),
...
 }
+
+#' @rdname MSnExperiment
+setMethod("spectrapply", "MSnExperiment", function(object, FUN = NULL,
+                                                   BPPARAM = bpparam(), ...) {
+    BPPARAM <- getBpParam(object, BPPARAM = BPPARAM)
+    isOK <- validateFeatureDataForOnDiskMSnExp(object@spectraData)
+    if (length(isOK))
+        stop(isOK)
+    file_f <- factor(object@spectraData$fileIdx,
```
sgibb: What do you think about an `fcol` argument? (see also https://github.com/lgatto/MSnbase/issues/422#issuecomment-459114666)

R/Backend-class.R
```diff
@@ -240,13 +217,16 @@ setMethod(
...
+#'     processing steps to be applied to the spectra before returning them.
 #' @return A list of [Spectrum-class] objects.
 #' @family Backend generics
 #' @author Sebastian Gibb \email{mail@@sebastiangibb.de}
 #' @noRd
 setGeneric(
     "backendReadSpectra",
-    def=function(object, file, spectraData, ..., BPPARAM=bpparam())
+    def=function(object, spectraData, processingQueue = list(), ...,
+                 BPPARAM=bpparam())
```
sgibb: I thought adding `BPPARAM` to all methods was a good idea but now we revert the design and `MSnExperiment,spectrapply` is doing the parallelisation. So maybe we should remove `BPPARAM` from `backendReadSpectra`/`backendWriteSpectra`. (I am not familiar with nested parallelisation.)

I would keep it for `backendImportData` to speed up backend creation.

What do you think is the best way?

R/BackendMzR-class.R
```diff
@@ -4,17 +4,17 @@ NULL
...
-    if (length(spd_list) != length(file))
-        stop("Number of files in 'spectraData' has to match length of 'file'")
-    res <- bpmapply(FUN = .spectra_from_file_mzR, file, spd_list,
-                    BPPARAM = BPPARAM, SIMPLIFY = FALSE)
-    names(res) <- NULL
-    res <- unlist(res, recursive = FALSE)
-    res[rownames(spectraData)]
+                                                       BPPARAM = bpparam()) {
+    file_f <- factor(spectraData$fileIdx, levels = unique(spectraData$fileIdx))
+    fls <- object@files[as.integer(levels(file_f))]
```
sgibb: Wouldn't be `fls <- object@files[spectraData$fileIdx]` enough, or if you are afraid of stupid (non-numeric) `fileIdx`:
```r
cfileIdx <- as.character(fileIdx)
fls <- object@files[match(cfileIdx, unique(cfileIdx))]
```
(should be a function)

R/AllClassUnions.R
sgibb: Ok, fine.

R/Backend-class.R
sgibb: Hasn't change in this PR preview (in the line above).

R/MSnExperiment-class.R
jotsetung: I like the principal idea. To keep it even more generic I would however then go for an argument `f = spectraData(object)$fileIdx`. This gives the user even more possibilities to define the split, without requiring the column to be in the `spectraData`. If we do that we **have** to ensure that the returned spectra are in the expected order!

R/Backend-class.R
jotsetung: nested paralellization: very bad thing. Yes, we should kick the parameter out of all `backend...` methods (except `backendImportData`).

R/BackendMzR-class.R
jotsetung: What I wanted to make sure here is that the `spectraData` is split how it is ordered, i.e. if (by whatever reason) `spectraData` contains first spectra from the second file and then from the first, that spectra from the second file are read first. I just wanted avoid to return the results as `res[rownames(spectraData)]`.

jotsetung: What would be the benefit from your solution?

sgibb: > What would be the benefit from your solution?

None. In fact it is the same as `factor` would do internally (but with more code etc.). I am sometimes confused about how `factor` work (ordered vs unsorted, index vs names etc.). With my solution I exactly see what happens. But it doesn't matter here, keep what you like more.

Issue #417, opened 2019-01-27
Add subset operator for Backend and its subclasses


sgibb:
- [x] `Backend` - @sgibb 
- [x] `BackendMemory` - @sgibb 
- [x] `BackendMzR` - @jotsetung 
- [x] `BackendHdf5` -

jotsetung:
You mean `[`? Another key subsetting operator here is `filterFile`.

sgibb:
Yep, I meant `[`, but you are right we also need `filterFile`. I would overload `[` with `i=fileIdx`, `j=spIdx`.
`filterFile,Backend` would be just `backend[fileIdx,]`.

I am wondering whether we should save the src file path in the backend instead of `MSnExperiment`. Because in `BackendMemory` can't reproduce the files from the `fileIdx` otherwise. And a subset method with `i, j, files, spectraData` argument looks awkward. 

A possible alternative would be `[i=srcFilename, j=spIdx]` (that wouldn't require src files in `Backend`).

But in fact it seems more natural that `Backend` stores the src file path instead of `MSnExperiment`, or?

jotsetung:
Yup. I agree. having the original files in the backend makes more sense.

jotsetung:
Question is what would make more sense, `[i, j]` with `i` being a file index and `j` a spectrum index or *vice versa*. From the usual way Bioconductor handles data matrices (columns samples, rows features) I would opt for `i` subsetting spectra and `j` subsetting files.

An implementation note for on-disk backends: subsetting by file should directly subset the `@files` slot, subsetting by spectrum would be optional (or just ignored, as for the `BackendMzR`).

sgibb:
You are right: `i=spIdx`, `j=fileIdx/fileName` would be more bioc-like. I had the filesystem/hdf5 approach (file->spectra) in mind. In fact it doesn't matter at all. Let's go for the bioc version.

jotsetung:
The default `[` for `Backend` simply subsets the `@files` slot based on `j`. For `BackendMzR` that is sufficient - for `BackendMemory` the spectrum index `i` should also be considered to reduce the internal list of spectra.

lgatto:
What is the motivation for changing the semantic of `[`? It used to be `i, j = "missing", drop = "missing"`, i.e. subsetting only on spectra. Also, `[spIdx, fileIdx]` uses the matrix/dataframe convention, but violates it, as it isn't rectangular. 
[...]

PR #424, opened 2019-02-01
Add filter/subsetting methods for MSnExperiment and Backend

jorainer:


R/Backend-class.R
```diff
@@ -249,3 +249,23 @@ setGeneric(
...
+#' @param i `integer`; ignored.
+#'
+#' @param j `integer` defining the files to which `x` should be subsetted.
+#'
+#' @return A `Backend` class.
+#'
+#' @author Johannes Rainer
+#'
+#' @rdname hidden_aliases
+setMethod("[", "Backend", function(x, i, j, ..., drop = TRUE) {
```
sgibb: See discussion in #417. Maybe we should not ignore `i` (global spectra index (file crossing)) but `j` and provide an additional method `filterFile`.

R/MSnExperiment-class.R
```diff
@@ -99,7 +129,44 @@ NULL
...
+#' ## Get associated file names
+#' fileNames(dta)
+#'
+#' ## Get spectra metadata
+#' spectraData(dta)
+#'
+#' ## Extract all spectra; by default a `Spectra` is returned. We could also
+#' ## get a simple `list` of `Spectrum` objects by specifying
+#' ## `return.type = "list"`.
+#' spctra <- spectra(dta)
```
sgibb: `spctra` seems to be a wired variable name in an example, IMHO even `s` or `sp` would be better (or remove all vowels: `spctr` :wink:)

R/MSnExperiment-class.R
```diff
@@ -303,10 +377,60 @@ spectraData <- function(object) {
...
+    fileNames(object@backend)
+})
+
 ##============================================================
 ##  --  SUBSETTING AND FILTERING METHODS
 ##
 ##------------------------------------------------------------
+#' @rdname MSnExperiment
+setMethod("[", "MSnExperiment", function(x, i, j, ..., drop = TRUE) {
+    if (!missing(i) & !missing(j))
```
sgibb: Also #417 .

R/utils.R
```diff
@@ -1499,3 +1499,22 @@ levelIndex <- function(x, which = c("first", "middle", "last")) {
...
+#'
+#' @author Johannes Rainer
+#'
+#' @noRd
+.to_index <- function(x, i, variable = "i") {
+    if (is.character(i))
+        i <- match(i, x)
+    if (is.logical(i))
+        i <- which(i)
+    if (any(is.na(i)) | any(i < 1) | any(i > length(x)))
```
sgibb: Please use `anyNA(i)` instead of `any(is.na(i))` and `||` instead of `|` (because we want to break ASOP, e.g. if any `i` is `NA` we don't want to test for `i < 1`):
```suggestion
    if (anyNA(i) || any(i < 1) || any(i > length(x)))
```

R/BackendHdf5-class.R
```diff
@@ -0,0 +1,149 @@
...
+    h5 <- H5Fopen(h5file)
+    comp_level <- .hdf5_compression_level()
+    fh <- openMSfile(file)
+    hdr <- header(fh)
+    pks <- peaks(fh)
+    close(fh)
+    for (i in seq_along(pks)) {
+        .pks <- pks[[i]]
+        colnames(.pks) <- c("mz", "intensity")
+        h5write(.pks, h5, paste0("spectra/", i), level = comp_level)
```
sgibb: I use the `H5Gopen` in my prototype to avoid creating this `group/id` string every time: https://gist.github.com/sgibb/54c1477d26fa8702231c447badd11776#file-class-draft-r-L198
But I am not sure that this is useful at all.

R/BackendHdf5-class.R
```diff
@@ -0,0 +1,149 @@
...
+.serialize_msfile_to_hdf5 <- function(file, h5file) {
+    h5 <- H5Fopen(h5file)
+    comp_level <- .hdf5_compression_level()
+    fh <- openMSfile(file)
+    hdr <- header(fh)
+    pks <- peaks(fh)
+    close(fh)
+    for (i in seq_along(pks)) {
+        .pks <- pks[[i]]
+        colnames(.pks) <- c("mz", "intensity")
```
sgibb: In my tests the colnames are not written to hdf5 (and we don't really need them; `mzR::peaks` don't has colnames either).

R/BackendHdf5-class.R
```diff
@@ -0,0 +1,149 @@
...
+#' Write the content of a single mzML/etc file to an h5file. We're using the
+#' spectrum index in the file as data set ID.
+#'
+#' @return `character(1)` with the md5 sum of the spectra data.
+#'
+#' @author Johannes Rainer
+#'
+#' @noRd
+.serialize_msfile_to_hdf5 <- function(file, h5file) {
+    h5 <- H5Fopen(h5file)
```
sgibb: I like the `on.exit(H5close(h5))` approach because it closes the file even if we have an error before the function is correctly finished.

R/Backend-class.R
sgibb: Please note: I just moved this myself to `filterFile` because I wanted it for the `BackendMemory` class.

R/MSnExperiment-class.R
jotsetung: Fixed.

R/Backend-class.R
jotsetung: I would not use `filterFile,Backend` and `[,Backend`. See https://github.com/lgatto/MSnbase/issues/417#issuecomment-460157900

R/MSnExperiment-class.R
jotsetung: It's gone. `[` just supports `i` now.

R/BackendHdf5-class.R
jotsetung: well, that makes life easier.

R/BackendHdf5-class.R
jotsetung: Could you do some performance tests on that?

R/BackendHdf5-class.R
```diff
@@ -0,0 +1,149 @@
...
+    if (length(x) != length(y))
+        "different number of md5 sums and hdf5 files"
+    else NULL
+}
+
+.valid.BackendHdf5.hdf5file <- function(x, y) {
+    msg <- NULL
+    if (length(x) != length(y))
+        msg <- "different number of hdf5 and original files"
+    if (length(x) != length(unique(x)))
```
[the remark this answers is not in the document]
jotsetung: Fixed now.

R/BackendHdf5-class.R
```diff
@@ -0,0 +1,149 @@
...
+                                                       ...) {
+    path <- suppressWarnings(normalizePath(path))
+    dir.create(path, showWarnings = FALSE, recursive = TRUE)
+    n_files <- length(files)
+    object@files <- files
+    object@md5sum <- character(n_files)
+    object@hdf5file <- character(n_files)
+    comp_level <- .hdf5_compression_level()
+    for (i in seq_len(n_files)) {
+        h5file <- paste0(path, "/", digest(files[i]), ".h5")
```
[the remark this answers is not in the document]
jotsetung: fixed.

R/BackendHdf5-class.R
sgibb: If we put the `paste0("spectra/", i)` out of the loop it makes no difference. I would suggest my slightly adapted version:
```r
.mzr2hdf5 <- function(file, h5) {
    msfh <- openMSfile(file)
    ## read complete header first to avoid random seq fault errors from
    ## proteowizard
    hdr <- header(msfh)
    on.exit(close(msfh))

    lvl <- .hdf5_compression_level()

    h5fh <- H5Fopen(h5)
    on.exit(H5Fclose(h5fh), add=TRUE)

    nms <- paste0("spectra/", seq_along(msfh))

    for (i in seq(along=msfh)) {
        h5write(peaks(msfh, i), h5fh, nms[i], level=lvl)
    }
}

jotsetung: totally agree. I did that already in an updated version.

PR #426, opened 2019-02-03
[WIP] Alternative BackendHdf5 implementation

sgibb:
@jotsetung this is an alternative but very similar implementation based on your suggestion (I mostly cherry-picked your previous work).

Unfortunately it is not tested (and won't work) in the current state. I am currently out of time but didn't want to waste your time. So please see it as WIP and suggestion. 

`backendReadSpectra`/`backendWriteSpectra` are not even tested by myself and unit tests are missing as well.

Main difference:
- it calculates checksums over files instead of whole spectra (because it is possible to update just a subset of spectra into a file which will not work with your approach. your code requires all spectra of a file in memory)


R/BackendHdf5-class.R
```diff
@@ -0,0 +1,225 @@
...
+    function(object, spectraData, ...) {
+
+    uFileIdx <- unique(spectraData$fileIdx)
+    h5files <- object@h5files[uFileIdx]
+
+    changed <- .valid.BackendHdf5.checksums(object@checksums[uFileIdx], h5files)
+    if (!is.null(changed))
+        stop(changed)
+
+    spectra <- vector(mode="list", length=nrow(spectraData))
```
jotsetung: Although I usually don't like having too many `if` conditions, I would make here a switch on whether data from a single or multiple files should be read:
```
if (length(h5files) == 1) {
  <code to read spectra from one file>
} else {
  <lapply/mapply over files>
}
```
The reason is that we want to make reading data as fast as possible and splitting, creating lists, assigning etc can slow stuff down. I tested that for the `BackendMzR` backend where I do also switch on the number of files `backendReadSpectra` should read data. Also, reading from multiple files should be the exception.

R/BackendHdf5-class.R
```diff
@@ -0,0 +1,225 @@
...
+
+setMethod(
+    "backendReadSpectra",
+    "BackendHdf5",
+    function(object, spectraData, ...) {
+
+    uFileIdx <- unique(spectraData$fileIdx)
+    h5files <- object@h5files[uFileIdx]
+
+    changed <- .valid.BackendHdf5.checksums(object@checksums[uFileIdx], h5files)
```
jotsetung: This is an important test. My only question is if it wouldn't be better to write the file's md5 sum into the hdf5 file (as entry `/md5/md5` as I have in my `BackendHdf5` implementation). How fast is creating the md5 sum of a hdf5 file in contrast to reading it from the file?
 
It's all about performance here. Would be sad if we have a fast backend, but slow it down because of slow data integrity tests.

R/BackendHdf5-class.R
```diff
@@ -0,0 +1,225 @@
...
+    spectra <- vector(mode="list", length=nrow(spectraData))
+    spectra <- split(spectra, spectraData$fileIdx)
+    spd <- split(spectraData, spectraData$fileIdx)
+    spIdx <- split(spectraData$spIdx, spectraData$fileIdx)
+
+    for (i in seq(along=h5files)) {
+        h5fh <- H5Fopen(h5files[i])
+        h5gfh <- H5Gopen(h5fh, "spectra")
+        spectra[[i]] <- .spectra_from_data(
+            lapply(as.character(spIdx[[i]]), h5read, file=h5gfh), spd[[i]]
```
jotsetung: Honestly, I wouldn't use `rhdf5::h5read`. While being the *official* exported function to read data from an Hdf5 file it is also slow if multiple small matrices (like in our case) are read (because of the many validations that are performed).

I spent a lot of time figuring out how to improve the speed of the backend in the *hdf5_jo* branch. We should use code from there (for which also unit tests exist).

R/BackendHdf5-class.R
jotsetung: I did a little comparison of md5 sums:
```r
sciex_h5 <- readMSnExperiment(sf, backend = BackendHdf5())
sps <- spectra(filterFile(sciex_mzr, 1))

fl1 <- sciex_h5@backend@h5files[1]
h5_fid <- .Call("_H5Fopen", sciex_h5@backend@h5files[2], 0L, PACKAGE = "rhdf5")

## If we read from the h5 file.
.read_checksum <- function(x) {
    fid <-.Call("_H5Fopen", x, 0L, PACKAGE = "rhdf5")
    on.exit(invisible(.Call("_H5Fclose", fid, PACKAGE = "rhdf5")))
    MSnbase:::.h5_read_bare(fid, "/md5/md5")
}

library(microbenchmark)
microbenchmark(digest::digest(sps),
                   digest::digest(sciex_h5@backend@h5files[1], file = TRUE),
                   .read_checksum(fl),
                   MSnbase:::.h5_read_bare(h5_fid, "/md5/md5"),
                   times = 5)
Unit: microseconds
                                                     expr        min         lq
                                      digest::digest(sps) 174483.582 181643.647
 digest::digest(sciex_h5@backend@h5files[1], file = TRUE)  43428.032  49818.164
                                       .read_checksum(fl)    271.155    350.599
              MSnbase:::.h5_read_bare(h5_fid, "/md5/md5")     45.878     51.455
        mean     median         uq        max neval cld
 194672.5498 193776.073 203786.325 219673.122     5   c
  54820.2748  51157.394  63760.690  65937.094     5  b 
    597.0336    512.348    664.490   1186.576     5 a  
    139.3908     76.466    166.528    356.627     5 a  
```

Summarizing, my way to calculate the checksum on the spectra is pretty bad, calculating checksums on the hdf5 file is faster - but reading the pre-calculated checksum from the hdf5 file beats everything. So I think we should go for that option - ideally on an already opened file.

R/BackendHdf5-class.R
```diff
@@ -0,0 +1,225 @@
...
+                collapse="\n"
+            )
+        ))
+    }
+    NULL
+}
+
+.valid.BackendHdf5.h5files <- function(src, h5files) {
+    if (length(src) != length(h5files))
+        return("Different number of hdf5 and original source files.")
```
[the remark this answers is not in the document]
sgibb: Ok, I guess that's a matter of taste. I can accept both.

R/BackendHdf5-class.R
sgibb: I called my slot `checksums` because I thought we don't want to use `md5` (I think collisions aren't a real problem for us). What about `xxhash32` or `crc32`? At least on my laptop they are 2-3x faster than `md5`.

The problem is that the hash sum calculation on the objects requires all objects in memory (even if you just modified one spectrum). 
There is `utils::fileSnapshot` and `utils::changedFile` that uses `file.info` (ctime, atime, mtime, size) to check whether a file changed (optionally md5 is supported).
We could also use the {a,c,m}time and size and just calculate the checksum if one of the times changed (if the size changed we already know it is different). That should be even faster without the need to store the checksum in the h5 file or to load all the objects into memory.

R/BackendHdf5-class.R
sgibb: Sorry, but I wasn't familiar with all your experiments/investigations you did before. But the speed gain you achieved is really impressive (10-15x on my laptop :+1::+1:)

sgibb:
> @sgibb , how should we proceed?

This PR was more or less to demonstrate my idea of the checksum approach. So I am closing it.

I think we should discuss `file.info` and recalculation of the checksums if needed.

PR #435, opened 2019-02-08
Add BackendHdf5,modCount; closes #429

### LINKED ISSUE #429: Use `file.info` instead of md5 sums to check for changed hdf5 files
Picking up the discussion in https://github.com/lgatto/MSnbase/pull/426#discussion_r253657248 wit @sgibb's excellent suggestion on the checksum.
We want the file checking to be really fast thus calculating the md5 sum on the file each time is slow. 

Question is when do we need to check for the hdf5 file to be changed.
- each time we read data from the file?
- in the `validObject`?

Following some tests I did over in the `hdf5_jo` branch I think it is crucial to check for changes each time data is read from the file. To check also in `validObject` is optional I think and not really necessary. What we want to check is if some other process changed the file or if the user made a copy of the object and changed data on that.

sgibb:
Replace `@checksums` by `@modCount`; see #429 for a detailed discussion.

### DIFF
--- a/R/BackendHdf5-class.R
+++ b/R/BackendHdf5-class.R
@@ -8,8 +8,16 @@ NULL
 #' @md
 #'
 #' @noRd
-setClass("BackendHdf5", contains = "Backend", slots = c(checksums = "character",
-                                                        h5files = "character"))
+setClass("BackendHdf5",
+    contains = "Backend",
+    slots = c(
+        # is incremented every time `backendWriteSpectra` is called and is used
+        # to test for changes of the hdf5 files and superficial copying
+        # see issue https://github.com/lgatto/MSnbase/issues/429
+        modCount = "integer",
+        h5files = "character"
+    )
+)
 
 #' @rdname hidden_aliases
 setMethod("show", "BackendHdf5", function(object) {
@@ -20,10 +28,11 @@ setMethod("show", "BackendHdf5", function(object) {
 #' @rdname Backend
 BackendHdf5 <- function() new("BackendHdf5")
 
-.valid.BackendHdf5.checksums <- function(x, y) {
+.valid.BackendHdf5.modCount <- function(x, y) {
     if (length(x) != length(y))
-        "different number of md5 sums and hdf5 files"
-    else NULL
+        "different number of hdf5 files and modification counter"
+    else
+        NULL
 }
 
 .valid.BackendHdf5.h5files <- function(x, y) {
@@ -43,7 +52,7 @@ BackendHdf5 <- function() new("BackendHdf5")
 }
 
 setValidity("BackendHdf5", function(object) {
-    msg <- c(.valid.BackendHdf5.checksums(object@files, object@checksums),
+    msg <- c(.valid.BackendHdf5.modCount(object@h5files, object@modCount),
              .valid.BackendHdf5.h5files(object@h5files, object@files),
              .valid.BackendHdf5.h5files.exist(object@h5files))
     if (length(msg)) msg
@@ -76,7 +85,7 @@ setMethod("backendInitialize", "BackendHdf5", function(object, files,
     dir.create(path, showWarnings = FALSE, recursive = TRUE)
     n_files <- length(files)
     object@files <- files
-    object@checksums <- character(n_files)
+    object@modCount <- integer(n_files)
     object@h5files <- character(n_files)
     comp_level <- .hdf5_compression_level()
     object@h5files <- file.path(path, paste0(.vdigest(files), ".h5"))
@@ -87,7 +96,7 @@ setMethod("backendInitialize", "BackendHdf5", function(object, files,
     for (i in seq_len(n_files)) {
         h5 <- H5Fcreate(object@h5files[i])
         h5createGroup(h5, "spectra")
-        h5createGroup(h5, "checksum")
+        h5createGroup(h5, "modification")
         H5Fclose(h5)
     }
     validObject(object)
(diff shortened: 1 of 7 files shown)


R/BackendHdf5-class.R
```diff
@@ -263,7 +270,7 @@ setMethod("backendReadSpectra", "BackendHdf5", function(object, spectraData,
 #' @md
 #'
 #' @noRd
-.h5_write_spectra <- function(x, spectraData, h5file, prune = TRUE) {
+.h5_write_spectra <- function(x, spectraData, h5file, modCount, prune = TRUE) {
```
sgibb: Is `prune = TRUE` a good idea? If we allow arbitrary splitting factors in `spectrapply` (and if we are automatically and immediately writing the data back to hdf) we could remove spectra we have written in a different thread, e.g.:
```r
# 1 file with 10 spectra
spectrapply(mse, f=rep(1:2, each=5), removePeaks)
# - first thread, write spectra, remove all spectra data and write spectra 1:5 to file 1
# - second thread, write spectra, remove all spectra data and write spectra 6:10 to file 1
#  (spectra 1:5 are now lost)
```

jotsetung: Writing data to hdf5 takes quite some time - that was one reason why I would not like to write every change immediately. And from a performance perspective it's much faster to remove the group within the file that contains all spectra and save all spectra in one go. Overwriting individual data sets does not work (if the dimension of the matrices differ). For now I would like to keep it that way.

sgibb: While I guess `h5delete(h5, "spectra")` is faster than `for (i in seq(along(x)) { h5delete(h5, name=spids[i]) }` the latter would IMHO be safer (and if you use again the raw internal API it should be relatively fast).

We could keep it in that way for now but we have to keep an eye on it.

sgibb:
I don't think we need `@modCount` in all backends (just in the modifiable backends with external (=non-R) dependencies like file or database backends). For `BackendMemory` it is meaningless because if we copy the object in R the spectra are copied as well. Currently just the `BackendHdf5` backend needs this counter.

> From BackendMemory or BackenHdf5 we can only convert to the BackendMzR if the counter is 0.

If we can't write to `BackendMzR` any conversion to it would be useless (especially if the counter is zero because than we just created the other backend from `BackendMzR`).

If you like we could move it to `Backend` and ignore it in `BackendMemory` and `BackendMzR`.

R/BackendHdf5-class.R
jotsetung: It depends on what you have in mind. For me, with the present code, it makes sense to always write all spectra of one file and never a single spectrum. And for that use case its saver to delete all content before. Imagine that you e.g. subsetted `spectraData` with `filterRt` before, now you have spectra in the hdf5 file that are no longer in `spectraData`.

In any rate, I agree that we might want to change this later if there is need.

jotsetung:
I still think the `@modCounter` would be useful in the `Backend` class for the following reason:

I would use as much as possible the lazy processing queue that we have in `MSnExperiment`. I would not by default call `backendWriteSpectra` each time that we modify or filter the object, but only if the user specifically asks for it (e.g. using a `consolidateData` or similar function). Each time `backendWriteSpectra` is called increments the counter in every backend - except `BackendMzR` (yes, I would also increment it in `BackendMemory`).

The counter can be used for both use cases: first for `BackendHdf5` (or any other writeable backend) to check if the data on disk has changed and secondly for the backend switching. Switching to a read-only backend only works if the counter is 0. And yes, I would allow switching from `BackendMemory` or `BackendHdf5` to `BackendMzR`. I don't see a reason why we should prevent that. And changing the backend from e.g. `BackendMemory` to `BackendMzR` is also different to creating a new `MSnExperiment` with `BackendMzR` backend (using the same files) from scratch, because it would keep the lazy processing queue.

sgibb:
> I would use as much as possible the lazy processing queue that we have in MSnExperiment.

Of course.

> I would not by default call backendWriteSpectra each time that we modify or filter the object, but only if the user specifically asks for it (e.g. using a consolidateData or similar function).

I thought `backendWriteSpectra` would be called automatically if the user calls `spectrapply` (which applies the whole processing queue, afterwards  we could write the changes back (for `BackendHdf5`), why keep them in memory?

But I have absolutely no problem to move the `@modCount` to `Backend`. I am going to do it the next days.

jotsetung:
> I thought `backendWriteSpectra` would be called automatically if the user calls `spectrapply` (which applies the whole processing queue, afterwards we could write the changes back (for `BackendHdf5`), why keep them in memory?

The idea of `spectrapply` was to apply a function to each spectrum and return its results. The lazy processing queue is applied on-the-fly to the data. I see that we will still have to discuss on when applying the processing queue and making its changes persistent. I thought of using the lazy processing queue as much as possible and making changes persistent only on a dedicated function call (see issue #440).

sgibb:
Ok, I moved the `@modCount` to the `Backend` class. IMHO we should provide a `isModified` (or `wasModified`) method that has to be implemented by all modifiable backends and will be called in `backendReadSpectra,Backends` to ensure that the backend wasn't modified (and to not duplicate the modification testing code in all `backendReadSpectra,*` methods).
I am going to implement `isModified` the next days.

jotsetung:
> IMHO we should provide a `isModified` (or `wasModified`) method that has to be implemented by all modifiable backends

Totally agree.

jotsetung:
Excellent @sgibb . Thanks for the changes.

Issue #440, opened 2019-02-11
Add a method that makes changes of the lazy processing queue permanent


jotsetung:
This picks up the discussion from issue #405. In summary, we agreed to use the lazy processing queue as standard for any data manipulations, i.e. `clean,MSnExperiment` adds the `clean` call to the object's lazy processing queue and applies it only (but each time) spectra data is retrieved.

To test this I've already implemented `removePeaks,MSnExperiment` and `clean,MSnExperiment` and it works nicely with all backends. `BackendMemory` and `BackendHdf5` would however allow to apply changes and write them back to the backend.

The worflow would be something like this:
```r
## Perform smoothing and peak picking
data <- smooth(data)
data <- pickPeaks(data)

## Both are in the lazy processing queue. If we're happy
## with both we could apply these changes to make them
## persistent
applyProcessingQueue(data)

## changes were written to the backend. The processing
## queue of `data` is now empty
```

Side note: we could even implement an `undo` method that removes the last processing step from the lazy processing queue - just an idea.

Name suggestions:
- `applyProcessingQueue`
- `consolidate`
- `force`

I wouldn't use `write...` because we have e.g. `writeMSData` to export the data to mzML (`exportMSData` would have been better).

@lgatto @sgibb other suggestions, preferences?

sgibb:
I like the `applyProcessingQueue` because it tells exactly what is going on but I would prefer to use
`data <- applyProcessingQueue(data)`. In `R` nobody is expecting side effects from a function and `applyProcessingQueue` will modify the backend and/or metadata and clean the processing queue (= really changing the whole object).

jotsetung:
I get your point @sgibb and agree. Only thing I wanted to avoid is that people use `data_2 <- applyProcessingQueue(data)`.

jotsetung:
I've implemented the following function:
```r
applyProcessingQueue <- function(x, f = spectraData(x)$fileIdx,
                                 BPPARAM = bpparam()) {
    if (!inherits(x, "MSnExperiment"))
        stop("'x' is supposed to be an 'MSnExperiment' object")
    if (length(x@processingQueue)) {
        isOK <- validateFeatureDataForOnDiskMSnExp(x@spectraData)
        if (length(isOK))
            stop(isOK)
        if (length(f) != nrow(spectraData(x)))
            stop("Length of 'f' has to match 'nrow(spectraData(x))'")
        ## Does not work on BackendMemory!
[...]

PR #443, opened 2019-02-20
Issue #440: add method to make changes persistent

jotsetung:
This pull request adds the method to make data manipulations persistent.

The reason to add yet another method to the backend was that I did not see another (clean) way to have different options to deal with applysing data manipulations: *in-memory*  backends simply apply the manipulations, but for some *on-disk* backends it is simply not possible to first load all spectra data, apply manipulations and then save it again. See issue #440 for more details.

With that method I believe we have all principal functionality in place.


jotsetung:
Agree.

jotsetung:
Especially since we already a related discussion on where to put the processing queue. Ideally, the `backend` should be agnostic of the lazy processing queue - only, I didn't find another way to implement a memory-efficient solution for ondisk backends (that don't have to load first all spectra, apply the processing queue to then save all spectra again).

jotsetung:
A fresh look at the code might indeed be good. Your suggestion sounds nice.

jotsetung:
@sgibb 's changes are now in this PR. Looks better now.

R/BackendHdf5-class.R
```diff
@@ -257,11 +257,11 @@ setMethod("backendReadSpectra", "BackendHdf5", function(object, spectraData,
     h5 <- H5Fopen(h5file)
     on.exit(invisible(H5Fclose(h5)))
     comp_level <- .hdf5_compression_level()
-    if (prune) {
-        h5delete(h5, "spectra")
-    }
-    if (!H5Lexists(h5, "spectra"))
-        h5createGroup(h5, "spectra")
+    x <- force(x)
```
sgibb: Do we really need `force` here? To be honest I never used it before.

R/MSnExperiment-class.R
```diff
@@ -479,53 +496,84 @@ readMSnExperiment <- function(file, sampleData, backend = BackendMzR(),
...
+        stop("Can not change backend to 'BackendMzR' because the ",
+             "data was changed.")
+    object@backend <- backendInitialize(backend, fileNames(object),
+                                        object@spectraData, ...)
+    validObject(object)
+    object
 })
 
+#' @rdname MSnExperiment
+applyProcessingQueue <- function(x, BPPARAM = bpparam()) {
```
sgibb: BTW: I know we had this discussion before but is there a specific reason why this is a function and not a method?

R/BackendHdf5-class.R
jotsetung: Yes! I had it without and got very weird errors. In fact it's R's lazy evaluation. I passed `x = spectra(object)` to the function and R was first deleting the file contents and then evaluating the `spectra(object)` expression - but the files were already empty. So, with `force` we make sure that we first read the spectra from the files **before** we delete their content.

sgibb: Ok, makes sense, thanks.

PR #448, opened 2019-02-28
Remove backendImportData

sgibb:
In https://github.com/lgatto/MSnbase/pull/441#issuecomment-463136543 @jotsetung mentioned that he
> would not mix using the backend to store the data and backend to be used as a data importer. 

This PR suggests just that. Now we have `setBackend` that allows switching backends. `readMSnExperiment` uses always `BackendMzR` to generate the `MSnExperiment`. Before it is returned we call `setBackend` and switch the backend to the one the user asked for.
This greatly reduce code complexity and lines without loosing anything.

There is only one problem: `setBackend` uses `backendWriteSpectra` which increments the `@modCount` by one so `setBackend(..., BackendMzR())` fails in the current implementation (that is the reason for the one failing unit test). IMHO we have two possible solutions:
1. Allow a `@modCount <= 1` in `setBackend,MSnExperiment,BackendMzR`.
2. Decrement the `@modCount` by one in `setBackend`.

I would vote for the first solution.

What do you think?

### DIFF
--- a/R/MSnExperiment-class.R
+++ b/R/MSnExperiment-class.R
@@ -487,7 +487,11 @@ readMSnExperiment <- function(file, sampleData, backend = BackendMzR(),
         metadata = metadata,
         processing = paste0("Data loaded [", date(), "]")
     )
-    setBackend(msnexp, backend, ..., BPPARAM = BPPARAM)
+
+    if (!inherits(backend, "BackendMzR"))
+        msnexp <- setBackend(msnexp, backend, ..., BPPARAM = BPPARAM)
+
+    msnexp
 }
 
 #' @rdname MSnExperiment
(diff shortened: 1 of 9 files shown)


R/MSnExperiment-class.R
```diff
@@ -477,19 +477,17 @@ readMSnExperiment <- function(file, sampleData, backend = BackendMzR(),
...
+                                smoothed = smoothed, BPPARAM = BPPARAM)))
+    msnexp <- new("MSnExperiment",
+        backend = backendInitialize(BackendMzR(), file),
         sampleData = sampleData,
         spectraData = spectraData,
         processingQueue = list(),
         metadata = metadata,
         processing = paste0("Data loaded [", date(), "]")
     )
+    setBackend(msnexp, backend, ..., BPPARAM = BPPARAM)
```
jotsetung: And what if we just use `setBackend` if the `backend` is not a. Sure, it adds another `if` condition, but I have a bad feeling with allowing `modCount` to be `1`. If you created an `MSnExperiment` with the constructor function you would get a `modCount = 0`.

Thus I would suggest:
```r
if (!inherits(backend, "BackendMzR"))
    setBackend(msnexp, backend, ..., BPPARAM = BPPARAM)
```

sgibb: That's better! I will change that.

sgibb:
IMHO the `setBackend` method should neither update nor reset the `@modCount` (because the data weren't changed). That's why `backendWriteSpectra` gains an `updateModCount` argument (that should never be `FALSE` except for `setBackend`).

jotsetung:
Yes, agree.

PR #484, opened 2019-10-03
Add parameter and change defaults for `consensusSpectrum`

jorainer:
- Add parameter `mzFun` to allow specifying the function to aggregate m/z values.
- Set default of parameter `mzd = 0` (was missing before which caused to estimate the parameter on the data - not ideal).
- Change default for parameter `intensityFun` from `max` to `median`. Reason: reporting the median intensity instead of the maximum was performing better on an example DDA data set (building consensus spectra from multiple MS2 spectra associated to the same precursor ion).


jorainer:
You're right @sgibb , they were accidentally added - I removed them again. Thanks for the review!

Issue #478, opened 2019-09-03
Add parameter msLevel to pickPeaks


jorainer:
For DDA data it might be helpful/better to use different settings for peak picking in MS1 and MSn levels (see http://www.metabolomics-forum.com/index.php?topic=1359.0).

We could add the `msLevel` parameter to perform peak picking specifically on a single MS level, or to use different settings for different MS levels.

Tony-II:
Hello jorainer,

I tried this recent implementation for DDA with MSnbase 2.11.6 and Biocondunctor 3.10, here especially **without** the smoothing (http://www.metabolomics-forum.com/index.php?topic=1359.0)

cent_file <- pickPeaks(x = raw_file, refineMz = "descendPeak", signalPercentage = 33, msLevel. = 1, verbose = TRUE)

I get:

Error in (function (classes, fdef, mtable)  :
  unable to find an inherited method for function ‘pickPeaks’ for signature ‘"missing"’

the raw file is a DDA-sample.
MSn experiment data ("OnDiskMSnExp")
Object size in memory: 2.67 Mb
- - - Spectra data - - -
 MS level(s): 1 2
 Number of spectra: 7205
 MSn retention times: 0:0 - 21:60 minutes
- - - Processing information - - -
Data loaded [Thu Sep 19 13:36:33 2019]
 MSnbase version: 2.11.6
- - - Meta data  - - -
phenoData
  rowNames:   xyz.mzML
  varLabels: sampleNames
  varMetadata: labelDescription
Loaded from:
  xyz.mzML
protocolData: none
featureData
  featureNames: F1.S0001 F1.S0002 ... F1.S7205 (7205 total)
  fvarLabels: fileIdx spIdx ... spectrum (35 total)
  fvarMetadata: labelDescription
experimentData: use 'experimentData(object)'

jorainer:
Would it be possible to share one of your files for me to try to reproduce? Unfortunately I don't have any profile-mode DDA files available.

Tony-II:
yes, I suggest to provide one DDA-file via a wetranfer download link. Would that be okay?

jorainer:
Whatever works for you is good for me.

Tony-II:
ok, here comes the link for one Demo file. The link is only valid for 7 days.
https://we.tl/t-VBD9NQYcaj

jorainer:
[...]

PR #485, opened 2019-10-14
feat: Add parameter msLevel. to smooth and fix setting of centroided variable

jorainer:
This addresses two things:
- add parameter `msLevel.` to `smooth` to allow smoothing of spectra from specific MS levels (issue #478).
- for `pickPeaks`, set only `centroided = TRUE` to spectra of the MS level specified with `msLevel.`.


jorainer:
Did remove the *NEWS* files - actually I don't have any pre-commit-news.sh file - I got the NEWS by the merge from master.

@lgatto, if you're also OK with this, could you merge please and push to BioC?

Issue #486, opened 2019-10-19
not properly reading fdata on some MRM samples


jmbadia:
Hello, 
I am reading a mzML file (attached: [27076.1.zip](https://github.com/lgatto/MSnbase/files/3746607/27076.1.zip)) with MSnBase acquired by MRM from QqQ samples. There are some chromatograms with identical polarity, precursor and product m/z values on it, but with different collision energies

> <chromatogram index="3" id="SRM SIC Q1=380 Q3=263.996 start=2.002616667 end=6.492633333"...>...<...name="collision energy" value="15.0">
> <chromatogram index="4" id="SRM SIC Q1=380 Q3=263.996 start=2.0044 end=6.494416667" ....>...<...name="collision energy" value="7.0">
> <chromatogram index="5" id="SRM SIC Q1=380 Q3=263.996 start=2.005283333 end=6.495316667"...>...<...name="collision energy" value="25.0"> 

I cannot read their feature Data correctly. fData() returns the data from the first chromatogram repeated three times, so I'm not able to know the fData from the others chromatograms (I need their collision energy, specifically)
```
Chroms <- readSRMData("27076.1.mzML")
fData(Chroms)[5:7,c("chromatogramId","precursorCollisionEnergy")]
```
> chromatogramId
> 5 SRM SIC Q1=380 Q3=263.996 start=2.002616667 end=6.492633333
> 6 SRM SIC Q1=380 Q3=263.996 start=2.002616667 end=6.492633333
> 7 SRM SIC Q1=380 Q3=263.996 start=2.002616667 end=6.492633333
>   precursorCollisionEnergy
> 5                       15
> 6                       15
> 7                       15

Thanks in advance

jorainer:
For this I have to dig a little deeper into the respective (C++) code in the `mzR` package. I guess there might be some problem in extracting an unique chromatogram ID from the *id* string in the chromatogram header in the mzML file.

jorainer:
So, `mzR` is reading the files correctly:
```r
> head(mzR::chromatogramHeader(fd))
                                               chromatogramId chromatogramIndex
1                                                         TIC                 1
2 SRM SIC Q1=300 Q3=263.996 start=2.006183333 end=6.496216667                 2
3 SRM SIC Q1=300 Q3=281.996 start=2.006183333 end=6.496216667                 3
4 SRM SIC Q1=380 Q3=263.996 start=2.002616667 end=6.492633333                 4
[...]

PR #487, opened 2019-10-23
fix: consider also collision energy in readSRMData

jorainer:
- Consider also the precursor collision energy in the generation of unique
  identifiers for each chromatogram across input files (fixes issue #486).
- Update relevant documentation.


lgatto:
Done and pushed to Bioc.

PR #500, opened 2020-04-10
Initial support for mzTab-M 2.0

sneumann:
Hi, my Easter present is an initial support for mzTab-M,
simply extending the list of supported sections and extending 
the slots in the MzTab class. Couldn't test locally, as my laptop 
generates a SEGV when installing, but that's probably my borked R installation. 
Will add improved testing soon. Yours, Steffen

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,7 @@
 
 ## Changes in 2.13.5
 - Nothing yet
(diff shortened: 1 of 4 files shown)


lgatto:
Thanks!

lgatto:
Add you as contributor.

Issue #501, opened 2020-04-13
Valid row/colnames after reading mzTab


sneumann:
Unusual bug report: when reading mzTab as in https://github.com/lgatto/MSnbase/blob/master/tests/testthat/test_MzTab.R#L9, I am getting *valid* names in e.g. 

```
> colnames(xx@SmallMolecules)
 [1] "SML_ID"                                "SMF_ID_REFS"                          
 [3] "database_identifier"                   "chemical_formula"                     
...
[17] "abundance_assay.4."                    "abundance_assay.5."                   
[19] "abundance_assay.6."                    "abundance_study_variable.1."          
[21] "abundance_variation_study_variable.1." "abundance_study_variable.2."          
[23] "abundance_variation_study_variable.2." "opt_global_Progenesis_identifier"     

> length(xx@SmallMolecules[,"abundance_assay.3."])  ## Works with [-operator
[1] 55

> length(xx@SmallMolecules$abundance_assay.3.)      ## Also works with $-operator
[1] 55
```

Which means I can't use 'em as-is when writing mzTab back out to file. 
I'd have to do some string-foo and search&replace numbers enclosed in '.',
and even that could go wrong under some conditions.

OTOH the metadata names are fine (actually: *illegal* in R):
```
> names(xx@Metadata)[1:6]
[1] "mzTab-ID"                   "software[1]"                "ms_run[1]-location"        
[4] "ms_run[1]-scan_polarity[1]" "ms_run[1]-format"           "ms_run[1]-id_format"  ...

> xx@Metadata$'assays[1]'   ## Can't access a list element by illegal name with $-operator:
NULL

> xx@Metadata[['assay[1]']] ## Need to access a list element by [[-operator:
[1] "3injections_inj1_POS"
```

=> I'd like to propose to accept and use the illegal row/col names when it comes to mzTab, 
and deal with it in the downstream code where that happens. 

Objections ? 

Yours, Steffen

lgatto:
The naming in mzTab is looking for issues - that `[1]`... really?

The reason why the names in the metadata are the way they are (illegal) is because they don't stem from a dataframe, are casted to a list and [reshaped](https://github.com/lgatto/MSnbase/blob/master/R/MzTab.R#L109-L114), thus preserving their illegality.

How would you go about to use illegal column names?

sneumann:
As shown above, the `[` operator works, it needs quoting anyway: `xx@Metadata[['assay[1]']]`
[...]

PR #502, opened 2020-04-13
Feature/mztabm

sneumann:
This PR adds 
1) a possibly breaking change where the special characters like space, [ and ] are *not* modified upon readin, see discussion in #501 
2) A write.mzTab method which writes an MzTab object back to disc.

Points for discussion:
1) Is there any objection to the `check.names=FALSE` change ? 
2) is the implementation of `write.mzTab()` as method OK ? 
3) Which "flavour" of mzTab do we want to write w.r.t. to number of columns in the different sections ? Should it be rectangular ? I think there are cases where we need to "trim" empty columns before creating the MzTab slots.

Yours, Steffen


lgatto:
Could you please send a PR per requested change. 

- The first one would be the change to `check.names = FALSE`, which I am happy with. With that PR in, it will be possible to check that there isn't anything that breaks. 
- As for `write.MzTab`, it should be `writeMzTab` and a simple function will do. 
- We try to adhere to [coding style](https://rformassspectrometry.github.io/RforMassSpectrometry/articles/RforMassSpectrometry.html#coding-style), which I am happy to help fix. But that will only be easy if you provide specific PRs (see first point above).

tests/testthat/test_MzTab.R
```diff
@@ -11,7 +11,7 @@ test_that("MzTab creation and accessors", {
     expect_null(show(xx))
     ## Accessors
     expect_is(metadata(xx), "list")
-    expect_identical(length(metadata(xx)), 64L)
+    expect_identical(length(metadata(xx)), 65L)
```
sneumann, line 14: Turns out that the previous implementation without `headers=FALSE` swallowed the first line of the MTD section. After fixing that there is one more metadata line.

lgatto:
Are there more PRs to come, or should I push to Bioconductor now?

sneumann:
There are no imminent PRs. What's left is a decision about the dimensionality of things. 
Some mzTab files have columns of empty tabs to the right of the actual data.frame. 
R happily imagines some names like "V8" "V9" for such columns. This needs to be discussed 
with mzTab folks, possibly me reading specs again. I would opt for some (optional ?)
trimming of the MzTab@SmallMolecules data.frame upon reading. 
But as said, that's not coming soon. Yours, Steffen

lgatto:
Pushed to Bioc.

Issue #505, opened 2020-04-20
error writeMgfData when using connections


ricoderks:
Hi,

I'am trying to use `writeMgfData` with a file connection. If I use this code:

```
con <- file(description = "test.mgf",
            open = "w")
writeMgfData(object = swath_spectrum[[1]],
             con = con)
close(con)
```

I get the following error: `Error in file(description = con, open = "at") : invalid 'description' argument`

And the full traceback:

```
Error in file(description = con, open = "at") : invalid 'description' argument
5. file(description = con, open = "at")
4. writeMgfDataFile(list(object), con = con, COM = COM, TITLE = TITLE, verbose = isMSnbaseVerbose())
3. .local(object, ...)
2. writeMgfData(object = swath_spectrum[[1]], con = con)
1. writeMgfData(object = swath_spectrum[[1]], con = con)
```

I'am also wondering why it says `open = "at"` while I specified `open = "w"`?

I would like to use connections, because later I would like to append more MSMS spectra to the same mgf ile.

Using R 3.6.3 and MSnbase 2.12.0.

Best regards,
Rico

ricoderks:
I just had a look at the code and I think I fixed it. I'll send a PR. I haven't done this that often so I hope I get it right.

lgatto:
Thank you for the PR.

PR #506, opened 2020-04-20
Only if con is character (filename) create a file connection.

ricoderks:
This should fixed issue #505.


lgatto:
Thank you. Merged, and will push to Bioc right away.

PR #512, opened 2020-05-05
Fix import of mzTab from MS-Dial

sneumann:
The particular flavor of MS-Dial causes `read.delim()` to try adding rownames, 
which are `SME` for all rows, which just fails. So enforce that rownames are 
simply enumerated numbers and not taken from file. My R is still borked, 
so I couldn't run R CMD check yet. Yours, Steffen

### DIFF
--- a/NAMESPACE
+++ b/NAMESPACE
@@ -121,6 +121,8 @@ export(MSnSet,
        ## MzTab
        MzTab,
        smallMolecules,
+       moleculeFeatures,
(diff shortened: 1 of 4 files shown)


sneumann:
Fixes the import of mzTab Flavour by MS-Dial, adds a unit test for that, 
exports the accessor functions in NAMESPACE and adds the mzTab-M citation 
to their manpage. Yours, Steffen

Issue #515, opened 2020-06-17
Rename Chromatograms?


jorainer:
Along the line of issue #514 , should we also rename the `Chromatograms` class? Maybe into `ChromatogramMatrix` because in fact it is organizing `Chromatogram` objects in a  2-dimentional matrix (and is extending `matrix`).

lgatto:
What about `MChromatograms` - the `M` for matrix of chromatograms and the leading M to stick to the same pattern and in #514.

jorainer:
I'm OK with that. After that I'll have to update also `xcms` - guess there are not many more packages using/extending directly classes from `MSnbase`.

lgatto:
I don't think so. But the good thing is that we are early in the release cycle, so plenty of time to brake dependencies.

Issue #514, opened 2020-06-17
Rename Spectra to MSpectra to avoid confusion with the Spectra package


jorainer:
Rename the `Spectra` class to `MSpectra` to avoid confusion with the `Spectra::Spectra` class.

PR #516, opened 2020-06-17
Rename Spectra and Chromatograms

jorainer:
- rename `Spectra` to `MSpectra` (issue #514)
- rename `Chromatograms` to `MChromatograms` (issue #515)


jorainer:
I don't quite understand why travis fails - all running smoothly locally.

jorainer:
OK, warnings fixed.

lgatto:
What about also renaming `Chromatogram` to `MChromatogram`? I know it isn't a matrix, but it would make the relation `Chromatogram` -> `MChromatograms` more explicit and it would also avoid ambiguity with `MSnbase::Chromatogram` and `Spectra::Chromatograms`.

jorainer:
Hm, I don't think that would be necessary. To be consistent we should then also rename `Spectrum`, `Spectrum1`, `Spectrum2` to `MSpectrum`, ...

lgatto:
There is no `Spectrum*` in `Spectra` (and there will never be).

jorainer:
But that's the same with `Chromatogram` - we will not have a `Chromatograms::Chromatogram` class. In general I'm OK with renaming `MSnbase::Chromatogram` to `MSnbase::MChromatogram`, I just wanted to do as little changes as possible, because these class renamings might cause downstream problems in other packages.

lgatto:
It isn't critical at this stage, I just want to avoid having to do another such refactoring another time. 

Regarding the `Spectrum` class, do you remember the discussion that lead to [this](https://github.com/lgatto/Spectrum). If this ever materialises, we might very well hit a clash with `MSnbase`. I thought we would stay on the safe side with anticipating anything, even unlikely, by renaming to `MChromatogram`.

lgatto:
I looked at the changes and am doing some local checks for the current PR. Will merge soon.

Issue #521, opened 2020-09-02
Smoothing of chromatogram in retention time domain: `combineSpectraMovingWindow`


breidan:
I am doing profiling of some biological extracts with LC-APCI-MS and my data is rather noisy. Running `xcms::findChromPeaks,CentWaveParam` over the unsmoothed data results in sometimes very irregular peak assignments. So I want to use `combineSpectraMovingWindow` to first smooth the chromatographic data first. I am aware of the issues with MSnExp and OnDiskMSnExp for further processing with `xcms`.

I am using MSnbase 2.12.0 (am currently limited to R3.6) and `combineSpectraMovingWindow` only accepts a `mzd=` argument. That leads to issues with either of the two extremes of the scan range, f.i. at the low end (m/z 100) a mzd=0.01 relates to 100 ppm accuracy while at the high end (m/z 1000) this would be 10 ppm. This is being measured on an Orbitrap and I have better than 5 ppm accuracy which I would like to maintain. 

Now, `meanMzInts` accepts a `ppm=` argument. Anyway that can be issued to  `combineSpectraMovingWindow` and if not, could that be implemented?

lgatto:
I'll leave @jorainer reply to this one.

jorainer:
Now that's a rather advanced question. Thanks for the thorough explanation @breidan . I will have a look at that and come back to you.

jorainer:
I've added parameter `ppm` to `combineSpectraMovingWindows` - you can install the updated version for `MSnbase` 2.12.0 with: `devtools::install_github("lgatto/MSnbase", ref = "RELEASE_3_10")`. The fix is also in PR #522.

Note that you **have** to specify `mzd = 0` if you only want to use the `ppm` - if you don't specify an `mzd` it gets automatically estimated on the data.

Please let me know if this works for you @breidan .

breidan:
@jorainer, thanks for the very fast resolution of this. Seems to be working fine.

PR #522, opened 2020-09-02
refactor: add parameter ppm to combineSpectraMovingWindow

jorainer:
- Add parameter `ppm` to `combineSpectraMovingWindow` (issue #521).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,8 @@
 # MSnbase 2.15
 
 ## Changes in 2.15.5
-- Nothing yet.
+- `combineSpectraMovingWindow` gains parameter `ppm` to allow definition of m/z
+  relative differences.
 
 ## Changes in 2.15.4
 - Add splitByFile,OnDiskMSnExp method to provide a more efficient splitting.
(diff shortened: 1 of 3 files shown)


lgatto:
I assume this needs to be pushed upstreams to `devel`, or is it to `RELEASE_3_10`?

jorainer:
Sorry for the late reply - this should go into the current Bioc devel (i.e. Bioc master branch). The fix for RELEASE_3_10 is in branch RELEASE_3_10

lgatto:
Shouldn't this be `RELEASE_3_11`?

lgatto:
I looks like everything is fine, just a version confusion, but waiting for you to confirm.

jorainer:
I added the fix only for the master branch and also for the *old* RELEASE_3_10 because the user from issue #521 could not install a more recent R/Bioc version. The RELEASE_3_10 fix was for him to test if it was working at all.

lgatto:
But I can't push to `RELEASE_3_10` any more, this is the old Bioconductor release version. If there's a change to backport to release, it must be 3.11.

lgatto:
So there isn't anything to push to any release branch.

lgatto:
If ok for you, I would prefer to delete the `RELEASE_3_*` branches here on github, to avoid any confusion.

jorainer:
Sure, no problem. Deleted the RELEASE_ branch

PR #523, opened 2020-09-14
feat: add extractSpectraData function

jorainer:
- Add `extractSpectraData` to extract all data from an `MSnExp`, `Spectrum` or
  `MSpectra` as a `DataFrame`. This data could then be passed to the `Spectra`
  constructor to create a `Spectra` from the original object (see also issue
  https://github.com/rformassspectrometry/Spectra/issues/123).


lgatto:
Quick question before the review: why call this `extractSpectraData()` and not `spectraData()` in line with the `Spectra` nomenclature?

jorainer:
Hm, good point. Well, mainly because we're actually extracting here also the m/z and intensity values and not *just* the `spectraData` - but I can change that, no problem.

lgatto:
No, fine, let's keep it then. The might be an extensive operation for large data, so maybe the name *extract* will avoid it to be called spuriously.

PR #535, opened 2020-12-07
Add several methods for Chromatogram and MChromatograms objects

jorainer:
This PR:
- fixes the failing unit test for `formatRt`.
- re-structures the `Chromatogram` and `MChromatograms` documentation.
- adds `normalize,Chromatogram` and `normalize,MChromatograms`.
- adds `filterIntensity,Chromatogram` and `filterIntensity,MChromatograms`.
- adds `align,Chromatogram` method.
- adds `c,MChromatograms` method.


R/AllGenerics.R
```diff
@@ -104,3 +104,4 @@ setGeneric("isolationWindowUpperMz", function(object, ...)
     standardGeneric("isolationWindowUpperMz"))
 setGeneric("filterPrecursorMz", function(object, ...)
     standardGeneric("filterPrecursorMz"))
+setGeneric("align", function(x, y, ...) standardGeneric("align"))
```
lgatto, line 107: 1. This should probably be added to `ProtGenerics`
2. What about a less generic name that *align*? Something like `alignRt()` to avoid conflicts with totally unrelated fields?

jorainer: Totally agree. So, let's go for `alignRt` then. I will make a PR against `ProtGenerics` with that method.

jorainer:
Regarding the failing unit test: building the vignette fails - but that seems to be unrelated to the new functionality. I would have fixed that, but am not familiar with the failing code...

R/AllGenerics.R
jorainer: PR: https://github.com/lgatto/ProtGenerics/pull/13

jorainer: Commit https://github.com/lgatto/MSnbase/pull/535/commits/a23c3b11bdc0ca106469351a8e83de40faf36b3b renames `align` to `alignRt` and imports the generic method from `ProtGenerics`.

lgatto:
Also pushed to Bioc.

PR #538, opened 2021-02-11
refactor: move ProcessingStep to ProtGenerics

jorainer:
Remove the `ProcessingStep` class and import it from `ProtGenerics` instead. Functionality is the same, but the version from `ProtGenerics` does no longer extend `Versioned`. I've checked if this creates any issues/problems with `OnDiskMSnExp` objects saved with older `MSnbase` version and it does not - all works fine.


lgatto:
Ok to push to Bioc too?

jorainer:
Yes please! And thanks!

lgatto:
Done - with a version bump.

PR #546, opened 2021-06-18
Add compareChromatograms method

jorainer:
Add a method `compareChromatograms` that allows to calculate a similarity score between two `Chromatogram` objects.


jorainer:
No idea where the error from the unit test is coming from - seems from some other part of MSnbase

jorainer:
Error fixed.

jorainer:
OK, that's it. Now I won't add any more stuff to the PR. Looking forward to your review @lgatto

R/methods-Chromatogram.R
```diff
@@ -471,3 +500,14 @@ setMethod("alignRt", signature = c(x = "Chromatogram", y = "Chromatogram"),
...
               .align_chromatogram(x = x, y = y, method = method, ...)
           })
+
+#' @rdname Chromatogram-class
+setMethod("compareChromatograms",
+          signature = c(x = "Chromatogram", y = "Chromatogram"),
+          function(x, y, ALIGNFUN = alignRt, ALIGNFUNARGS = list(),
+                   FUN = cor, FUNARGS = list(use = "pairwise.complete.obs"),
+                   ...) {
+              if(length(x) != length(y) || !all(rtime(x) == rtime(y)))
```
sgibb, line 510:
```suggestion
              if (length(x) != length(y) || !all(rtime(x) == rtime(y)))
```

R/methods-MChromatograms.R
```diff
@@ -635,3 +645,55 @@ setMethod("alignRt", signature = c(x = "MChromatograms", y = "Chromatogram"),
...
+    }
+    nx <- length(x)
+    ny <- length(y)
+
+    m <- matrix(NA_real_, nrow = nx, ncol = ny)
+    for (i in seq_len(nx)) {
+        for (j in seq_len(ny)) {
+            if (!full && i > j)
+                next
+            m[i, j] <- compareChromatograms(x[[i]], y[[j]], ALIGNFUN = ALIGNFUN,
```
sgibb, line 693: The lower triangle of the comparison matrix would be a mirror of the upper one, or? So it is a waste of time to calculate everything twice (if `full == TRUE`).
Why you don't use the same logic as in [`Spectra:::.compare_spectra_self`](https://github.com/rformassspectrometry/Spectra/blob/70fc21d0c35726d0c7e09e21f233dc5ab31fe7f2/R/Spectra-functions.R#L333-L358)

(Is `full` then even needed?)

sgibb (changes requested):
I would suggest to not calculate the comparison metric twice.

R/methods-MChromatograms.R
jorainer: I would also guess the lower triangle to be the same than the upper triangle - so, I'll drop `full` and will calculate only the upper one (and the diagonal)

jorainer:
I addressed your comments @sgibb - no idea why the CI fails, locally all tests were OK.

R/methods-MChromatograms.R
jorainer: In the updated version only the upper triangle is calculated and then copied to the lower

sgibb (approved):
Great!

R/AllGenerics.R
```diff
@@ -1,6 +1,7 @@
 setGeneric("bin", function(object, ...) standardGeneric("bin"))
 setGeneric("clean", function(object, ...) standardGeneric("clean"))
 setGeneric("compareSpectra", function(object1, object2, ...) standardGeneric("compareSpectra"))
+setGeneric("compareChromatograms", function(x, y, ...) standardGeneric("compareChromatograms"))
```
lgatto, line 4: Seeing these generics, I am wondering if `compareSpectra()`, `compareChromatograms()` and probably other shouldn't be moved to `ProtGenerics`. The latter is defined in `MSnbase` and `Spectra` already. 
```r
> MSnbase::compareSpectra
standardGeneric for "compareSpectra" defined from package "MSnbase"

function (object1, object2, ...) 
standardGeneric("compareSpectra")
<bytecode: 0x563351276228>
<environment: 0x563350fb3438>
Methods may be defined for arguments: object1, object2
Use  showMethods(compareSpectra)  for currently available ones.
> Spectra::compareSpectra
standardGeneric for "compareSpectra" defined from package "Spectra"

function (x, y, ...) 
standardGeneric("compareSpectra")
<bytecode: 0x563347e2f818>
<environment: 0x563347e21d20>
Methods may be defined for arguments: x, y
Use  showMethods(compareSpectra)  for currently available ones.
```

lgatto: Also, I would suggest to name the arguments as in `Spectra`.

jorainer: Very good point @lgatto . I will add `compareSpectra`, `compareChromatograms` and also `bin` to `ProtGenerics` - they are all implemented also in `Spectra`. Also, I agree on a common way to name parameters, but I would change from `object1, object2` to `x, y` as we have it in `Spectra`.

ProtGenerics PR will come.

jorainer: ProtGenerics PR: https://github.com/lgatto/ProtGenerics/pull/15

jorainer:
the recent commit finished the transfer of the `bin`, `compareChromatograms` and `compareSpectra` methods to `ProtGenerics`.

lgatto:
FYI - The GHA error seems to originate from [MsCoreUtils](https://github.com/rformassspectrometry/QFeatures/pull/139#issuecomment-872860658).

lgatto:
Pushed to Bioc.

Issue #549, opened 2021-08-10
XIC without data in one sample


hvaret:
Hi,

I open this issue for a possible bug when using the `plot(..., type="XIC")` function. When analyzing my LCMS data I wanted to plot the XIC of several samples using a specific m/z-RT area, and it happened there was no data in this area for at least one of them. The strange behavior is that I actually obtained a plot for all the samples.

Below is a toy reproducible example using the data from the msdata package:

```
library(msdata)
library(MSnbase)
od <- readMSData(dir(system.file("sciex", package = "msdata"),
                     full.names = TRUE), mode = "onDisk")
# centroiding
od <- pickPeaks(od)
# data for both samples between mz=106.04 and 106.06
serine <- filterMz(filterRt(od, rt = c(175, 190)), mz = c(106.04, 106.06))
plot(serine, type = "XIC")
# no data between mz=106.053 and 106.056 for the second sample to reproduce the bug
serine <- filterMz(filterRt(od, rt = c(175, 190)), mz = c(106.053, 106.056))
plot(serine, type = "XIC")
```

Note that a plot is produced for the first sample (as expected) but not for the second one (error) when we try separately:

```
serine <- filterMz(filterRt(od[which(fromFile(od) == 1)], rt = c(175, 190)), mz = c(106.053, 106.056))
plot(serine, type = "XIC")
serine <- filterMz(filterRt(od[which(fromFile(od) == 2)], rt = c(175, 190)), mz = c(106.053, 106.056))
plot(serine, type = "XIC")
```

When looking at the source code my feeling is that the error is in https://github.com/lgatto/MSnbase/blob/80e3b08bff43dea8bd11e86737589254f10babbf/R/functions-MSnExp.R#L623 with `x` and `fns` that have different lengths.

Here is my `sessionInfo()`:

```
R version 4.1.0 (2021-05-18)
Platform: x86_64-apple-darwin17.0 (64-bit)
Running under: macOS Big Sur 11.5.1

Matrix products: default
LAPACK: /Library/Frameworks/R.framework/Versions/4.1/Resources/lib/libRlapack.dylib

locale:
[1] fr_FR.UTF-8/fr_FR.UTF-8/fr_FR.UTF-8/C/fr_FR.UTF-8/fr_FR.UTF-8

attached base packages:
[1] stats4    parallel  stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[...]

PR #550, opened 2021-08-16
fix: empty XIC plot if no data (issue #549)

jorainer:
This fixes issue #549: an empty plot is created if no data points are available for a file.


lgatto (approved):
Looks good to me, thanks.

lgatto:
@jorainer - this has apparently not been merged?

jorainer:
Ah, no, that's true. Can you please merge and push to BioC (I think I have no access to the MSnbase BioC repo)?

lgatto:
But some checks fail at the moment.

jorainer:
It's a very strange error that I get in the centroiding vignette only when running either `R CMD build MSnbase` or `R CMD check MSnbase`. The error does not happen when rendering the vignette with `rmarkdown` alone.

jorainer:
Forget my previous comment. It was a very stupid error. I called the `splitByFile,MSnExp` method in the wrong way. It was working on `OnDiskMSnExp` objects, but not on `MSnExp` objects. Now it should be fine.

lgatto:
Pushed to Bioc.

PR #557, opened 2021-11-08
Enable "fast load" also on macOS

jorainer:
Reading the header information for the last spectrum before importing the peaks matrix led to random errors on recent macOS/R versions - although this fix was introduces years back to avoid exactly those errors. This PR changes the default on macOS - Windows and linux installation are unaffected (always used *fast load = TRUE*).


jorainer:
Maybe it would also be good to include this fix also in the BioC 3.14 branch...

lgatto (approved):

lgatto:
Yes, will also backport to latest release.

PR #569, opened 2022-01-27
fix: adapt to mzR version 2.29.3

jorainer:
- mzR >= 2.29.3 drops support for Ramp, this the respective backend is no longer
  available in mzR.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: MSnbase
 Title: Base Functions and Classes for Mass Spectrometry and Proteomics
-Version: 2.21.4
+Version: 2.21.5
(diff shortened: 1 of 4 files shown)


lgatto (approved):

lgatto:
Shall I push to Bioc?

jorainer:
Yes, please. That should also fix eventual errors on the Bioconductor build machines, because the mzR update is already pushed to BioC.

lgatto:
done

lgatto:
Full strike ! https://master.bioconductor.org/checkResults/3.15/bioc-LATEST/MSnbase/

jorainer:
Should be fixed then with version 2.21.5 you submitted. That's still the old version I assume

PR #571, opened 2022-02-09
Allow different orientation of axis labels in image2

lmsimp:
In this PR I have added two new arguments to `image2` to allow users to set the `las` graphical parameter to change the orientation of axis labels. By default the x-axis labels are now perpendicular to the axes.

previous -
image (1)

new default -
image (2)

### DIFF
--- a/R/methods-MSnSet.R
+++ b/R/methods-MSnSet.R
@@ -450,25 +450,29 @@ image2 <- function(x,
                    y.cex.axis = .75,
                    xlab = "Samples",
                    ylab = "Features",
+                   x.las = 2,
+                   y.las = 0,
                    ...) {
-    if (inherits(x, "MSnSet"))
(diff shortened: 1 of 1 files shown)


lmsimp:
Related to the above, I also noticed that the default text sizes for the labels in `image2` is. `x.cex.axis = .75` and `y.cex.axis = .75,`. This is fine and imagine a smaller `cex` is needed as often we have many labels. However, I see that `naplot` which creates a heat map of missing values using `image2` and also a barplot above it (as per the above) both have different `cex.axis` for each y-axis. I'm not sure if this intentional. I tend to think it looks messy.  I think perhaps it is best to set the axis labels for the barplot also to .75?

lgatto:
Thanks. Merged, but you are missing a version bump and the NEWS file. Will do it now, but keep an eye on these next time.

Issue #583, opened 2022-10-28
Possible bug in descendPeaks


Pascallio:
Dear developers,

I might have encountered a bug for an unlikely, but possible situation in `MSnbase:::descendPeaks`. According to the documentation, "the peak region is defined by descending from the identified centroid/peak on both sides until the measured signal increases again.". However, in situations where the centroid is either the first or last signal in a spectrum, the other side of the peak seems to be ignored. 

Some example code for reproducibility:

```{r}
mz <- c(100.14, 100.145, 100.15) # masses
intensity <- c(1000, 700, 600) # intensities
peakIdx <- 1 # position of the centroid

MSnbase:::descendPeak(mz = mz, intensity = intensity, peakIdx = peakIdx) 
```

This results in a mass of 100.14 and does not take other values than the centroid into account (which in this example would have resulted in 100.1441).  It seems that `MSnbase:::descendPeak` calculates the `half_width`, which in this case results in 0 and therefore only considers the centroid itself.

Is this the correct behaviour or should the remaining signals be taken into account until the (first) observed signal rises again?

Best,
Pascal

jorainer:
Dear Pascal! Thanks for the pointing this out! Would you be willing to provide a pull request that fixes this issue?

Pascallio:
Hi Johannes, 

Yes ofcourse, I've just written a fix and provided the pull request.

Cheers,
Pascal

lgatto:
I have merged and pushed to Bioc in devel and release. Thank you very much @Pascallio !

PR #586, opened 2022-11-20
Fix for Issue #583

Pascallio:
This fix adds the parameter `descendIfEdge`, which changes the behavior of `descendPeak` if set to TRUE. Doing so will continue the descend if the centroid is near the end of the spectrum by using the full width instead of the half width. Defaults to FALSE to keep the current behavior to preserve current centroid results.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -51,6 +51,10 @@ Authors@R: c(person(given = "Laurent", family = "Gatto",
              person(given = "Miguel", family = "Cosenza-Contreras",
                     email = "migueljcc5@gmail.com",
                     role = "ctb"),
+             person(given = "Pascal", family = "Maas",
+                    email = "p.maas@lacdr.leidenuniv.nl",
+                    comment = c(ORCID = "0000-0001-9379-6041"),
+                    role = "ctb"),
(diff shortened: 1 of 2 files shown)


jorainer (approved):
Looks good to me @Pascallio !

@lgatto , can you please merge (if you're also OK with the PR) bump version and push to BioC?

jorainer:
@Pascallio can I also ask you to have a look whether also the (newer) function in MsCoreUtils has the same issue?

https://github.com/rformassspectrometry/MsCoreUtils/blob/master/R/refineCentroids.R

lgatto:
Thank you. @Pascallio, if you want, feel free to add yourself to the contributors in the DESCRIPTION file. Bug reports and fixes are important contributions to open source software and deserve proper credit.

Pascallio:
Thanks @lgatto , I will add myself to the DESCRIPTION file.

It seems the refineCentroids function does not have this exact issue @jorainer , however I did find two other issues (possibly related to this issue) for which I will open separate issues in MsCoreUtils.

On a related note, I made an RCPP version of the descendPeaks function that I can share if you are interested?

PR #593, opened 2023-06-24
Fix declarations of centroided/smoothed setters for OnDiskMSnExp objects

hpages:
Salut Laurent,

With current R-devel (R 4.4), trying to install **MSnbase** <= 2.25.2 fails with error messages complaining about the declarations of the `centroided<-` and `smoothed<-` methods for OnDiskMSnExp objects. The complain is:

    arguments ('value') after '...' in the generic must appear in the method, in the same place at the end of the argument list

We've been notified by R-core that this breaks many CRAN packages that depend directly or indirectly on **MSnbase** when using the latest R devel.

This commit fixes the method delcarations so that they are compatible with R >= 4.4.

Best,
H.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: MSnbase
 Title: Base Functions and Classes for Mass Spectrometry and Proteomics
-Version: 2.25.2
+Version: 2.25.3
 Description: MSnbase provides infrastructure for manipulation,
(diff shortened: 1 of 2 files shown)


lgatto:
Merci beacoup! And sorry for the hassle.

lgatto:
I have pushed this to devel on Bioc too.

hpages:
Merci Laurent !

PR #600, opened 2024-03-13
refactor: move XML to Suggests

jorainer:
This PR moves the XML package from Imports to Suggests and removes the import from the NAMESPACE. Package version was not bumped.


lgatto (approved):

jorainer:
thanks for checking. I leave merging and pushing to BioC (and bumping version) to you @lgatto

lgatto:
done

PR #605, opened 2024-09-12
feat: add function to coerce from Spectra to MSpectra

jorainer:
- Add support to allow coercing from a `Spectra` to a `MSpectra` object using the `as()` method.
- Update documentation to describe conversion between `Spectra` and `MSpectra`.


jorainer:
Note: unit tests are OK - GHA failed because of codecov.

lgatto:
Sorry, missed the deadline. Will be in 2.33.1.

PR #609, opened 2025-04-10
chore: remove R/methods-fragments.R (moved to PSMatch)

sgibb:
After adapting to the new `PSMatch::calculateFragments` (with `fixed_modifications` instead of `modifications`) some unit tests in `topdownr` fail. In `topdownr` `MSnbase::calculateFragments` overwrites `PSMatch::calculateFragments`.
Unfortunately the syntax is slightly different, e.g. `MSnbase` knows nothing about `fixed_modifications`.

To reproduce this error, load `MSnbase` before `PSMatch`:

```r
library("MSnbase")
library("PSMatch")
calculateFragments("AACE", fixed_modifications=NULL)
# Error in .local(sequence, ...) : 
#   unused argument (fixed_modifications = NULL)
```

`MSnbase::calculateFragments`'s body is similar to `PSMatch::calculateFragments`'s body. The only difference is the `verbose = isMsnbaseVerbose()` argument/setting.
So maybe it would be useful to remove this file? (or adjust to `PSMatch`'s arguments).


lgatto:
Indeed. I am surprised that there's still such a function in `MSnbase`.

lgatto (approved):

PR #614, opened 2026-08-07
refactor: use selective imports from BiocGenerics

jorainer:
- Move ProtGenerics and BiocGenerics from Depends to Imports
- Use selective imports from BiocGenerics.


NAMESPACE
```diff
@@ -1,12 +1,29 @@
...
+importMethodsFrom("BiocGenerics", "mad")
+importMethodsFrom("BiocGenerics", "rownames")
+importMethodsFrom("BiocGenerics", "rownames<-")
+importMethodsFrom("BiocGenerics", "sapply")
+importMethodsFrom("BiocGenerics", "scale")
+importMethodsFrom("BiocGenerics", "unsplit")
+importMethodsFrom("BiocGenerics", "updateObject")
+importMethodsFrom("BiocGenerics", "width")
+
+import("ProtGenerics")
```
lgatto, line 22: Why is there the full `import("ProtGenerics")` and the selective `importMethodsFrom()` below?

jorainer: Originally it was just importing full ProtGenerics. I started adding selective imports, but then did not proceed - it took me already quite some time to do them for BiocGenerics.

